PR #581, opened 2014-12-30
HLE boot / HLE bios

skmp:
So, this is working for the most part. I don't think it's ready for shipping, but I guess since quite a few games do run it might be a useful default?

I'd say around 10-15% of the games work right now, depends on the devkit version they used. Most early games boot (HH, recv, msr, msvc, hotd, g2, etc), and some seem to run fine. Most later sdk versions seem to be broken. Some games do weird stuff (i'm looking at you, ouga).


skmp:
Merging this as it doesn't break android, but haven't tested that it works.

We have to re-factor the java side some time soon...

PR #585, opened 2015-01-09
http support for cdi & gdi, updated based gdipsr

skmp:
- Fix coreio fsize
- Fix coreio fopen to seek to 0
- Fix coreio/http to smart-escape urls
-  For urls that include \ or % in the filenames, you have to pass them
   escaped
- Update gdi parser to use (mostly) streamstream. That code is horrible
  and should be rewritten
- coreio core is hacky and horrible at places as well
- Update imgreader to use coreio
- Update cdi parser + driver to use coreio

### DIFF
--- a/core/deps/chdpsr/cdipsr.cpp
+++ b/core/deps/chdpsr/cdipsr.cpp
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
+#include "deps/coreio/coreio.h"
 #include "cdipsr.h"
 
 // Global variables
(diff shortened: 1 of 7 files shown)


skmp:
The gdipsr code isn't as strict and allows single line gdi files (matches \n or space or tabs).
I'm not sure if this is an issue. I'm also not sure of scanf strictly matches \n\r either.

Also, max gdi index file limited to 8K chars. I don't think this is an issue, but thinking on *very* large gdi descriptors ( ~ 100 tracks is max), the per-track filenames need to be < 50 chars. Almost all gdis I've seen use track%02d.ext, and the longest I've seen is 30smth tracks, so not likely to be an issue.

I need to stop writing horrible, hacky, one-hour-is-enough-to-add-this-feature style of code.

LoungeKatt:
The best stuff starts that way :)

PR #596, opened 2015-01-25
Add theme selection GUI, Flush incoming intent when handled

LoungeKatt:


skmp:
Hmm, after selecting a theme I can no longer use the default one. Also, looks for pngs in subfolders but then selecting them no longer works.

LoungeKatt:
I don't doubt it's very broken. This is the first time I've had time for anything but class and family issues in a while.

LoungeKatt:
isGame requires the leanback support library, which is incorrectly dependent on 4.4.2 and needs to be redirected to 5.0.1

I have a feeling this can't be built with Travis until the Android SDK is fixed.

PR #633, opened 2015-04-10
Fix german translations

Holzhaus:
- This fixes typos and other mistakes in the german translations - mostly the so-called "Deppenleerzeichen" (wrongfully inserted spaces in composed words). Refer to http://de.wikipedia.org/wiki/Leerzeichen_in_Komposita for details.
- This also adds german translations for all strings in `shell/android/res/values/strings.xml`.
- The ordering of the strings now corresponds to `shell/android/res/values/strings.xml`.


LoungeKatt:
Human translation is always better than Google.

Issue #455, opened 2014-05-02
Licensing violation, GPL, future plans & related discussion


skmp:
Mhrf, I hoped not having to deal with this for a little bit longer, but I guess shit happens   (#452)

I was already thinking of changing/discussing the license stuff for a while, I guess the time has come.
# Some legalese

In order for me to take action against this (and other violations) I need to be the main copyright owner. Copyright violations can't be pursued by 3rd parties.
## Current reicast copyright status

Here's the current copyright structure of reicast
- Parts I've written independently (this covers over 95% of the source) are owned by nillware, which is a company I own in Greece. 
- Parts that were developed by me under contract, for which I don't have the copyright, but are licensed as GPL
- Parts developed by the rest of the nulldc team (ZeZu, PsyMan). This code is under gpl, but the copyright belongs to them.
- 3rd party contributions to nulldc, before 2008-2009. We use them under GPL. (I don't know if there are any actually, but I have to check).
- Contributions to the reicast project (excluding mine). Copyright belongs to the contributors, we use them via GPL.
- 3rd party libraries. I believe most of these are under bsd/mit/or some other liberal license (but I'll have to cross check)
## Taking actions on violations

As nillware owns a very large part of the project, I can take legal action via this route, under Greek Law. If any of you want to you can also take legal action as your copyrights might also be violated (if your code was included in the happy chick build). The fact that the violation was definitely in "bad faith" and couldn't happen "accidentally" make this a fairly clear case for the US judicial system.

Any copyright owner can also request a take-down under the DMCA system.

As always, it's probably best to contact the author directly and take lega/dmca action only if such actions don't resolve this issue. We should do this once we decide on how we'll move forward.
# Future plans

Now, with that said, I'd really like to clean up the copyright status, and license reicast under a different license. 
## Why ?
- GPL and mostly restricts us (for example, it might be illegal to use the moga libraries), but we get very few benefits
- GPL in practice is almost never enforced, it's expensive to go after each and every violation that originates a continent or two away, and its a process that takes many years.
- I don't actually mind people re-using reicast that much, even in closed source scenarios. I'm not that much of an open sauce zealot. There has to be some form of attribution though.
- It might enable some source-level monetization, which IMO is _much_ nicer and more aligned with my ethics than selling it on a app store. I've put at least 5-6 full-time years into this project, and others have contributed significantly, with the expectation of it being free for the end user.
## What are my(nillware's :p) plans for the project?

[...]

PR #642, opened 2015-04-15
Audio backends cleanup

Holzhaus:
These commits clean up the audio backends and is an attempt to consolidate audio-specific code. This also makes it possible to have multiple audio backends compiled into reicast (e.g. ALSA & OSS on Linux). These can then be selected by the user via a config option ("backend" in the "audio" section). It *should* also be possible to reduce the number of platform-specific ifdefs with this approach.

Audio backends need to define a struct (`audiobackend_t`) that holds a pointer to it's init, push and term function and also a slug and a name as string. They then need to pass a pointer to this struct to `RegisterAudioBackend()` before `InitAudio()` is called.

For now, I defined a function (`RegisterAllAudioBackends()`) that get's called exactly once (i.e. upon first call of `InitAudio()`), where we can put `#ifdef`'d `RegisterAudioBackend()` calls into for each backend that reicast supports.

I subsequently added backends for OSS, ALSA, DirectSound, Android and PulseAudio. Please note that I wasn't able to test the DirectSound or Android backend, since I only have Linux. The DirectSound backend should work since I took most of the code from `ds_audiostream.cpp`, but I have no idea if the android backend is implemented correctly since I do not have any experience with Android programming in combination with native C/C++ code.

Although this compiles fine on my Linux x86_64, I actually had to apply commits a579466 to af64d37 from the [skmp/linux-x64](https://github.com/reicast/reicast-emulator/compare/skmp/linux-x64) branch to actually get sound at all.

Please check the individual commit messages for details.


skmp:
Long month, late reviews.

This does look good. I'll get it in after #655 is done

skmp:
hey @Holzhaus, can you take a look at #455 and add your name there if you agree? Then I can merge this ~

Holzhaus:
[Yup.](https://github.com/reicast/reicast-emulator/issues/455#issuecomment-103088650)

Holzhaus:
Should I rebase?

LoungeKatt:
After you do, I will merge this (as long as nobody objects). It sounds like @skmp approved and Its been in my local copy through many tests of other changes.

Holzhaus:
There you go. I recompiled and sound worked fine for me (on Linux x86_64), even without borrowing the commits a579466 to af64d37 from the skmp/linux-x64 branch.

Issue #645, opened 2015-05-07
The scheduling system is awful, rewrite it


skmp:
Was meant as a hack to test the speed benefits of the idea, not to be actually used. Recent commits should have fixed most of the brokage, but it's a golden example of perfectly broken design and implementation.

PR #646, opened 2015-05-08
tmu: Correctly schedule next interrupt in underflow, correctly differ…

### LINKED ISSUE #594: Frame rate limit and audio are broken too
Compared to the version on the Play Store, the latest CI build doesn't cap the framerate properly (it runs about twice as fast, and the frame limit option seems to do nothing) and the audio is a broken mess. What happened?

LG Tribute, quad-core 1.2 GHz, Android 4.4.2.

skmp:
…entiate between large values and underflows

- 32 bit code was "fudgy" on this, so simplified via 64bit counters for edge detection
- Maths to count cycles were wrong
- Let's not do 32 bit wrapped maths after beers without any kind of pen-and-paper design

Should fix #594, also see #645

### DIFF
--- a/core/hw/sh4/modules/tmu.cpp
+++ b/core/hw/sh4/modules/tmu.cpp
@@ -19,6 +19,7 @@ u32 tmu_prescaler_mask[3];
 
 u32 tmu_shift[3];
 u32 tmu_mask[3];
+u64 tmu_mask64[3];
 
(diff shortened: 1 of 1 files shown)


skmp:
http://reicast-builds.s3.amazonaws.com/builds/heads/fix/tmu-undeflow-logic-3fe8f0e/reicast-android-debug-3fe8f0e.apk

PR #652, opened 2015-05-09
Feat/no direct memmap

skmp:
This is a quick hack to verify that the 5.0 compat / random compat problems we have are caused by the direct mem mapping code.

So far I've had multiple independent confirmations that this resolves both new and old breakage.

This *must absolutely not* be merged in it's current state. I'll clean it up for r7 timeframe, then properly implement it for r8/r9.

### DIFF
--- a/core/hw/mem/_vmem.cpp
+++ b/core/hw/mem/_vmem.cpp
@@ -399,6 +399,8 @@ void _vmem_term()
 
 }
 
+bool _nvmem_direct_map = false;
+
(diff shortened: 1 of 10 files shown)


skmp:
HHmaay, my next goal is cleaning and merging this one in. *Crosses fingers*

skmp:
Implemented in 71c2cb4512d91f14fd22383514548391c9fa64b1 via #743

PR #671, opened 2015-06-19
linux: print memory map to stderr when fixed path not available

imirkin:
Not sure why it's dying on start yet for me, but it was doubly-mysterious when it exited instead of dumping some useful info.

### DIFF
--- a/core/linux/common.cpp
+++ b/core/linux/common.cpp
@@ -203,7 +203,11 @@ void print_mem_addr()
     if (ofp == NULL) {
         fprintf(stderr, "Can't open output file %s!\n",
                 outputFilename);
+#if HOST_OS == OS_LINUX
+        ofp = stderr;
+#else
         exit(1);
+#endif
     }
 
     char line [ 512 ];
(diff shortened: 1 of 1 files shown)


imirkin:
Was it foolish of me to assume that core/linux == not android? Is there an example of how to do this properly? Just #ifdef _ANDROID or something more subtle?

LoungeKatt:
If you are asking if basic linux and Android are the same, they are very similar on a kernel level. On a linux machine, you are likely running the emulator via a terminal, where it will crash back to the terminal and the error output will have printed to the terminal. On Android, you are writing the log to logcat, which is printed to a text file, and crashing back to the launcher. The exit terminates the emulator, which prevents a crash with somewhat misleading information (SEGFAULT, for example) from printing shortly after the memory log and wasting resources.

What you would want it something like

```#if HOST_OS == OS_LINUX
// don't exit, do something else
#else
exit(0);
#endif```

imirkin:
I had just meant the directory layout -- the file I was changing was core/linux/common.cpp. Sorry, I'm largely unfamiliar with the source layout at this point. I'll update the pull req as you suggested.

LoungeKatt:
The directory layout is almost identical in the system and very similar in storage. Android just has a few (annoying) eccentricities.

PR #673, opened 2015-06-23
Fixes for building iOS / Android in parallel

LoungeKatt:


core/nullDC.cpp
```diff
@@ -206,7 +206,11 @@ int dc_init(int argc,wchar* argv[])
 	
 	mem_map_default();
 
+#ifndef _ANDROID
```
skmp: I don't quite get the point of this #ifdef

core/hw/maple/maple_cfg.h
```diff
@@ -62,5 +62,9 @@ struct IMapleConfigMap
 	virtual ~IMapleConfigMap() {}
 };
 
+#ifndef _ANDROID
```
skmp: I don't quite get the point of this #ifdef

core/nullDC.cpp
LoungeKatt: "Prep for Linux" which prevents that pull request from breaking Android.

core/hw/maple/maple_cfg.h
LoungeKatt: "Prep for Linux" which prevents that pull request from breaking Android.

PR #675, opened 2015-06-27
Linux version loads controller information from the configuration file

bobbyblues:
For all versions:
 - The virtual controllers and VMU plugged are now read from the config file. If nothing is specified in the configuration file, the first controller with two VMUs are plugged by default.

For linux version:
 - Keyboard binding for each controller is read from the configuration file, using X11 keycodes
 - PC Controller mapping for each dreamcast controller is read from the configuration file as well

I have tested my code with two players using 1 keyboard for two players, 1 keyboard + 1 controller and 2 controllers, and everythings was working properly. Though I could not try using twice the same controller (which would then appear having the same name), but I think it should not be an issue.

Don't hesitate to tell me if you want me to change things before accepting the pull request.

### DIFF
--- a/core/hw/maple/maple_cfg.cpp
+++ b/core/hw/maple/maple_cfg.cpp
@@ -4,7 +4,6 @@
 #include "maple_devs.h"
 #include "maple_cfg.h"
 
-#define HAS_VMU
 /*
 bus_x=0{p0=1{config};p1=2{config};config;}
 Plugins:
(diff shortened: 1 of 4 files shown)


LoungeKatt:
By all versions, does this include Android (which no longer uses a configuration file at all)?

LoungeKatt:
A quick rebase of this should line it up with the code that was added to keep all the Android GUIs working (or until it is updated to use this method, too)

skmp:
@bobbyblues @Holzhaus As with #670 i'm closing this in favor of more recent work in #769. Would be great if changes could be integrated into the new PR.

PR #677, opened 2015-06-28
[PoC] Unified logging API

Holzhaus:
**This is not ready for merging, it's just a proof-of-concept implementation.**

I thought I would be cool to have a unified logging API, so that logging messages are `printf`'d on Unix/Windows, but send to the native logger on Android.

This is just a quick example implementation.

Please let me know If you find something like this useful.

### DIFF
--- a/core/linux-dist/main.cpp
+++ b/core/linux-dist/main.cpp
@@ -1,5 +1,6 @@
 #include "types.h"
 #include "cfg/cfg.h"
+#include "oslib/logging.h"
 
 #if HOST_OS==OS_LINUX
 #include <poll.h>
@@ -178,13 +179,13 @@ void SetupInput()
 		ioctl(JoyFD,JSIOCGBUTTONS,&ButtonCount);
 		ioctl(JoyFD,JSIOCGNAME(sizeof(Name)),&Name);
 		
-		printf("SDK: Found '%s' joystick with %d axis and %d buttons\n",Name,AxisCount,ButtonCount);
+		log_msg(LOG_INFO, "sdk", "Found '%s' joystick with %d axis and %d buttons", Name, AxisCount, ButtonCount);
 
 		if (strcmp(Name,"Microsoft X-Box 360 pad")==0)
 		{
 			JMapBtn=JMapBtn_360;
 			JMapAxis=JMapAxis_360;
-			printf("Using Xbox 360 map\n");
+			log_msg(LOG_INFO, "input", "Using Xbox 360 map");
 		}
 	}
 #endif
@@ -284,7 +285,7 @@ bool HandleKb(u32 port) {
 	return true;
 	#else
   	while(read(kbfd,&ie,sizeof(ie))==sizeof(ie)) {
-		printf("type %i key %i state %i\n", ie.type, ie.code, ie.value);
+		log_msg(LOG_DEBUG, "input", "type %i key %i state %i\n", ie.type, ie.code, ie.value);
 	}
 	#endif
 #endif
@@ -549,7 +550,7 @@ void os_CreateWindow()
 			x11Display = XOpenDisplay( 0 );
 			if (!x11Display && !(x11Display = XOpenDisplay( ":0" )))
 			{
-				printf("Error: Unable to open X display\n");
+				log_msg(LOG_ERROR, "x11", "Unable to open X display");
 				return;
 			}
 			x11Screen = XDefaultScreen( x11Display );
@@ -825,7 +826,7 @@ int main(int argc, wchar* argv[])
 		x11_keymap[36] = Btn_Start;
 	#endif
 
-	printf("Home dir is: %s\n",GetPath("/").c_str());
+	log_msg(LOG_INFO, "env", "Home dir is: %s",GetPath("/").c_str());
 
 	common_linux_setup();
 
--- a/core/oslib/logging.cpp
+++ b/core/oslib/logging.cpp
@@ -1,0 +1,90 @@
+#include <stdio.h>
+#include <cstdarg>
+#include "oslib/logging.h"
+#if ANDROID
+#include <android/log.h>
+#endif
+
+#define LOG_ERROR   0
+#define LOG_WARNING 1
+#define LOG_INFO    2
+#define LOG_DEBUG   3
+#define LOG_VERBOSE 4
+
+#if ANDROID
+static void log_msg_android(int loglevel, const char *tag, const char *fmt, va_list args)
+{
+    /* This function sends messages to the native Android logging API.
+     * See http://mobilepearls.com/labs/native-android-api/#logging for details.
+     */
+
+    // Map our custom loglevel to the respective ANDROID_LOG_* level
+    int android_loglevel;
+    switch(loglevel)
+    {
+        case LOG_ERROR:
+            android_loglevel = ANDROID_LOG_ERROR;
+            break;
+        case LOG_WARNING:
+            android_loglevel = ANDROID_LOG_WARNING;
+            break;
+        case LOG_INFO:
+            android_loglevel = ANDROID_LOG_INFO;
+            break;
+        case LOG_DEBUG:
+            android_loglevel = ANDROID_LOG_DEBUG;
+            break;
+        case LOG_VERBOSE:
+            android_loglevel = ANDROID_LOG_VERBOSE;
+            break;
+        default:
+            android_loglevel = ANDROID_LOG_UNKNOWN;
+    }
+    // Then, send the message to the Android log
+    __android_log_vprint(android_loglevel, tag, fmt, args);
+}
+#endif
+
+static void log_msg_printf(int loglevel, const char *tag, const char *fmt, va_list args)
+{
+    const char *loglevel_str;
+    switch(loglevel)
+    {
+        case LOG_ERROR:
+            loglevel_str = "ERROR";
+            break;
+        case LOG_WARNING:
+            loglevel_str = "WARNING";
+            break;
+        case LOG_INFO:
+            loglevel_str = "INFO";
+            break;
+        case LOG_DEBUG:
+            loglevel_str = "DEBUG";
+            break;
+        case LOG_VERBOSE:
+            loglevel_str = "VERBOSE";
+            break;
+        default:
+            loglevel_str = "UNKNOWN";
+    }
+    // Print the tag and loglevel
+    printf("[%s][%s] ", loglevel_str, tag);
+
+    // Print the actual message
+    vprintf(fmt, args);
+
+    // Print newline
+    printf("\n");
+}
+
+void log_msg(int loglevel, const char *tag, const char *fmt, ...) {
+    va_list args;
+    va_start(args, fmt);
+#if ANDROID
+    log_msg_android(loglevel, tag, fmt, args);
+#else
+    log_msg_printf(loglevel, tag, fmt, args);
+#endif
+    va_end(args);
+}
--- a/core/oslib/logging.h
+++ b/core/oslib/logging.h
@@ -1,0 +1,9 @@
+#pragma once
+
+#define LOG_ERROR   0
+#define LOG_WARNING 1
+#define LOG_INFO    2
+#define LOG_DEBUG   3
+#define LOG_VERBOSE 4
+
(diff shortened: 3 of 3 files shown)


skmp:
I'd rather have less verbose functions/macros, eg, log_x11e, log_x11i, etc. In a lot of places, #ifdefing out the logs gives a big speed bump too (these places have #define wrappers around per-file log macros, like printf_gdrom) and we can have this work with more specific macros.

LoungeKatt:
This would unify the logs that are used, although it may not make much difference. The current method we use of pushing core errors through the JNI already logs these items to logcat on Android. This may end up being redundant, since allowing the core to handle the errors appropriately is a little more user-friendly (ie. presenting a popup instead of simply crashing).

Holzhaus:
@skmp Well, the naming can be changed of course. Perhaps, we could just put something like this in `logging.h`:
```C++
#ifdef LOGGING
#define LOG_E(tag, message, ...) log_msg(LOG_ERROR, tag, message,  __VA_ARGS__)
#define LOG_W(tag, message, ...) log_msg(LOG_WARNING, tag, message,  __VA_ARGS__)
#define LOG_I(tag, message, ...) log_msg(LOG_INFO, tag, message,  __VA_ARGS__)
#define LOG_D(tag, message, ...) log_msg(LOG_DEBUG, tag, message,  __VA_ARGS__)
#define LOG_V(tag, message, ...) log_msg(LOG_VERBOSE, tag, message,  __VA_ARGS__)
#else
#define LOG_E(tag, message, ...)
#define LOG_W(tag, message, ...)
#define LOG_I(tag, message, ...)
#define LOG_D(tag, message, ...)
#define LOG_V(tag, message, ...)
#endif
```

This way, all logging calls are stripped to boost performance if we do not define `LOGGING`, but we do not need `#ifdef`s everywhere (IMO the goal should be to have no `#ifdef`s at all in the core code). This could even be tuned so that we can define a logging level at compilation time (i.e. log errors and warnings, but not the other stuff).

If we use macros like this, logging should not cause a performance hit if we don't need it.

LoungeKatt:
@Holzhaus How does disabling logs to reduce the performance hit from a unified system provide a better solution than reducing the performance hit of the current logs without removing them?

Holzhaus:
It doesn't, but If I understood @skmp correctly, the current logging system also works with `#ifdef`fing out logging to gain performance.

Also, this simple logging system should not pose a severe performance hit. The `switch` statement should have a complexity of O(1) (branch table). Also, we could get rid of the `printf("\n");`, so all we do is a `printf` and a `vprintf`.

LoungeKatt:
So instead of platform ifdef's we would define and undefine LOGGING all over?

Holzhaus:
Nope, only once in the Makefile. We then use log level ´DEBUG´ or ´VERBOSE´ for logging calls that should not be used during normal operation and are only there for debugging purposes.

LoungeKatt:
So you still need platform ifdef's on top of it to disable specific logs. That sounds like doubling the performance hit.

skmp:
@Holzhaus that's almost exactly what I meant. By #ifdefs I meant on the header file.

Current code is like
[...]
void nilprintf(...){}
[...]
#define printf_rm nilprintf
#define printf_ata nilprintf

To "scope" the printfs and to enable me to control sensitive ones at compile time.

In many cases, just having a reference to printf (which means, a literal pointer to a string to the current constant pool and a possible call affecting register allocation), even if not run, can have 10-15% performance drop (Particularly in TA Fifo side of things). Same thing goes for many of the asserts I have all over, disabling them makes the jit process almost twice as fast. 

I would prefer to further cut this down to include the TAG, as I hate typing, and strings suck as identifiers. Something like

```
extern void vlog(log_level_t log_level, log_tag_t log_tag, const char* fmt, const va_list& vl);

#define LOG_FN(level,tag) static inline log_#level_#tag(const char* fmt, ...) { va_list vl; va_start(vl, fmt); vlog(LOG_LEVEL_#level, LOG_TAG_#tag, fmt, vl); va_end(vl); } \

#define LOG_FNS(tag) LOG_FN(e, tag) LOG_FN(w, tag) LOG_FN(i, tag) LOG_FN(d, tag) LOG_FN(d, tag) LOG_FN(v, tag)

LOG_FNS(x11)
LOG_FNS(gdrom)
LOG_FNS(ta)

log_w_ta("My message");
log_e_ta("another message");

//etc
```

LoungeKatt:
I guess it had nothing to do with the platform ifdefs. In that case, I will leave you two to sort this out.

Holzhaus:
@skmp Any particular reason why you don't like strings as identifiers? As far as I can see, we'd save two chars `"..."` at the cost of readability (thus maintainability) and flexibility.

As a compromise, how about something like this:

```C++
typedef const char* log_tag_t;

extern log_tag_t log_tag_x11 = "x11";
extern log_tag_t log_tag_gdrom = "gdrom";
extern log_tag_t log_tag_ta = "ta";

#define LOG_E(tag, ...) log_msg(LOG_ERROR, log_tag_##tag, __VA_ARGS__)

// An actual call look like this:
int main(void)
{
    LOG_E(x11, "This is a log message");
    LOG_E(x11, "This is a %s log message", "formatted");
    return 0;
}
```

We'd still save the quote chars, but we don't need (num_loglevels * num_tags) different macros.

skmp:
I don't like strings because they are extremely easy to mistype, and hard to enumerate. I also love getting code autocompletion, and dislike "mostly-same" parameters. Because of global namespace, we can't just use some very common names like x11, gdrom, pvr, ta, etc, without running into possible conflict and confusion (why is gdrom is dedicated to say, logging, and not register state?). Lastly, it is hard to find all calls to a function - log_e(x11, log_e (x11, log_e( x11 are all valid, thus you need regexprs - with a simple grep/search. 

You find the actual implementation hard to read (the macros?) or the usage?

skmp:
Oh also, I'm not sure all of our compile targets support the __VA_ARGS__ macro (that's why I defined static inline functions instead)

LoungeKatt:
@skmp The code you posted isn't obvious. It isn't complicated, but it shouldn't leave any questions as to how a log is printed.

Unity - Debug.Log ("content");
Android - Log.X("TAG", "content");
Reicast - Umm, log_X_ta("content"); ? Is my code broken or did I print the log wrong?

Holzhaus:
> I don't like strings because they are extremely easy to mistype, and hard to enumerate. I also love getting code autocompletion, and dislike "mostly-same" parameters.

I can't really argue against these points, but I don't consider them as huge problems since most modern editors/IDE are supporting custom snippets that should take care of that.

> Because of global namespace, we can't just use some very common names like x11, gdrom, pvr, ta, etc, without running into possible conflict and confusion (why is gdrom is dedicated to say, logging, and not register state?).

In case you're referring to my latest suggestion: `x11` is not a variable and thus does not conflict with the global namespace, since `x11` will be macro-expanded to `log_tag_x11` (which is a variable but shouldn't cause any conflicts or confusion). Strings won't have this problem either.

>it is hard to find all calls to a function - log_e(x11, log_e (x11, log_e( x11 are all valid, thus you need regexprs - with a simple grep/search.

IMHO, the proper way to deal with this is to enforce a coding style standard. In a python project I maintain, we're using a linter (`flake8`) and Travis-CI to make sure that PRs and new commits are auto-checked for coding style compliance. Also, I'm using a git pre-commit hook to cancel a commit if I didn't adhere to the coding standards. Maybe we can do that here, too?

>  I'm not sure all of our compile targets support the VA_ARGS macro (that's why I defined static inline functions instead). NDK used to lag very much behind that stuff, and I'm not sure about the apple side, either.

As for the portability concerns you posted, I tried to make sure that the code is portable.  `__VA_ARGS__` is part of the C99 standard, so they should support it. And both seem to do, according to these StackOverflow questions: [Android NDK](http://stackoverflow.com/questions/29557649/android-ndk-variadic-macro-requiring-1-args), [Apple-GCC](http://stackoverflow.com/questions/6750512/gcc-warning-iso-c-does-not-permit-named-variadic-macros)

The main portability issue is the usage of variadic macros without the variadric argument, e.g.:
```C
#define PRINT(fmt, ...) printf(fmt, __VA_ARGS__)

PRINT("this is a %s test", "working");
// Expands to:
printf("this is a %s test", "working");

PRINT("this is a test");
// Expands to:
printf("this is a test", ); // Note the trailing comma, doesn't work
```
You can circumvent this by using `##__VA_ARGS__ ` instead of `__VA_ARGS__` in GCC, which removes the trailing comma if no variadic arguments are present. However, this is not part of the C-standard (C11 perhaps?) and doesn't work with compilers other than GCC. MSVC works in all cases since it doesn't comply to the C standard by silently dropping the comma if there are no variadic arguments.

I'd just circumvent this whole issue by using:
```C
// Just consider fmt to be part of __VA_ARGS__
#define PRINT(...) printf(__VA_ARGS__)

PRINT("this is a %s test", "working");
// Expands to:
printf("this is a %s test", "working");

PRINT("this is a test");
// Expands to:
printf("this is a test"); // No trailing comma since fmt is part of __VA_ARGS__
```

> You find the actual implementation hard to read (the macros?) or the usage?

The macros/inline function mix. I guess all of the code including stuff like `LOG_FNS(x11)` go into `logging.h`?

If so, the usage should be only slightly more complicated than my initial proposition (i.e. "you need to add a new tag to logging.h before you can use it" in contrast to "just use any string you want as a tag in the code").

I still prefer [this approach](https://github.com/reicast/reicast-emulator/pull/677#issuecomment-118670590), but it's my code so I'm biased ;-)

As you may have noticed I enjoy discussion, but in the end I have to admit that you're the main developer, so you'll probably be the one who'll write a lot more logging calls than me. So basically it's your opinion that matters here and if you dislike strings as tag/identifiers so much, then we should go with your solution instead.

LoungeKatt:
@Holzhaus you make a good point, but it is good to speak up when something seems a bit confusing. It beats having to go back later and explain (for example) why you didn't add any logs to your code.

skmp:
Macros also live in the global namespace, so not much difference there.

```
#define x11 Blargh;
void* x11; // still breaks
```

I guess you refer to the linux linking issues, in which case you are right (I somehow never run into this problem).

> If so, the usage should be only slightly more complicated than my initial proposition (i.e. "you need to add a new tag to logging.h before you can use it" in contrast to "just use any string you want as a tag in the code").

That's was 80% of my reason.

Yeah, all of this logic goes to logging.h.

We're not coding against C99 though, we're coding against C++nooneknowshwatversion, usually somewhere between C++89 and C++14. Android indeed no longer uses outdated compiler versions though, so we're good there. 

I'd still like to have separate functions. I'll make a PR w/ it.

Holzhaus:
> Macros also live in the global namespace, so not much difference there.
>
> #define x11 Blargh;
> void* x11; // still breaks

You dont define `x11`. Try out this code, it compiles and works as espected:
```C
#include <stdio.h>

typedef const char* log_tag_t;

log_tag_t log_tag_x11 = "X11-Component";

#define PRINT(tag) printf("The tag is: %s\n", log_tag_##tag);

int main(void)
{
    const char *x11 = "this is a test\n";
    PRINT(x11);
    printf(x11);
    return 0;
}
```

Holzhaus:
@skmp @LoungeKatt Any news on this? I'd like to have some kind of logging with tags, that's also possible to turn off. The current output is a pain to read IMO.

I'd implement something, but we need to have a shared opinon on "pass tag as string" vs "dedicated logging function for every component or cpp file".

LoungeKatt:
As long as there is somewhere I can find all the tags to put into the log writer for Android, it is up to you two how you go about it.

skmp:
@Holzhaus you still around these days? I'll tryr to take a peek on this over the weekend

Holzhaus:
Yup, although I'm currently traveling so I can't make any code changes in the next few days.

Am 27. März 2018 10:37:54 GMT-05:00 schrieb Stefanos Kornilios Mitsis Poiitidis <notifications@github.com>:
>@Holzhaus you still around these days? I'll tryr to take a peek on this
>over the weekend

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

skmp:
(Cleaning up old tickets)

PR #680, opened 2015-07-03
Fix shm_open on the rpi

joolswills:
Wasn't actually putting the returncode of shm_open to fd

This code could do with a cleanup though -  as it doesn't need to fallback anymore is the fallback code needed ? fd doesn't need to be initialised either. Not sure if this branch is still in development or not really, so I'll leave that to you. Cheers.

### DIFF
--- a/core/hw/mem/_vmem.cpp
+++ b/core/hw/mem/_vmem.cpp
@@ -550,7 +550,7 @@ error:
 		//shm is borken on the rapi
 		//using a file is slower, but works
 		//file in tmpfs would be better?
-		fd = -1;shm_open("/dcnzorz_mem", O_CREAT | O_EXCL | O_RDWR,S_IREAD | S_IWRITE);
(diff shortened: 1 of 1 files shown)


skmp:
I think this was on purpose to disable the shm logic on rpi for that branch. I'll have to check.

joolswills:
the last commit was to re-enable it, but incorrectly. it works with this on the rpi. See 14875df52b547351d37d050fbd4e6f0154b343cb

```
commit 14875df52b547351d37d050fbd4e6f0154b343cb
Author: Stefanos Kornilios Mitsis Poiitidis <skmp@nillware.com>
Date:   Wed Feb 25 20:59:47 2015 +0100

    linux/dist: shm w/ noexec + READ_IMPLIES_EXEC + mprotect bug + personality fix = shm works on rapi2
```

skmp:
Oh, that makes sense. I need to properly merge this to master, but till then I'll have this merged. Read on mind #455 for licensing

PR #684, opened 2015-07-05
Close all open popups from hardware key press

LoungeKatt:


LoungeKatt:
The menu still doesn't close using hardware buttons, which contradicts the three different attempts to dismiss the popups.

PR #694, opened 2015-07-21
Cleanup

randomstuff:


core/nullDC.cpp
```diff
@@ -196,7 +197,6 @@ int dc_init(int argc,wchar* argv[])
 		printf("Using Interpreter\n");
 	}
 	
-	  void InitAudio();
```
skmp: I'm just too damned lazy sometimes -.-

skmp:
I'm favor apart from the verify changes -- the code should be changed to be like rbi != 0 (and the ==false left in place). Maybe i'm nitpicking. Can you rebase to latest master?

randomstuff:
Yes, your solution for `verify` is better…

randomstuff:
Rebased and fixed the `verify` macro usage.

skmp:
Does this also address #718 ?

PR #695, opened 2015-07-21
cmake build (currently Linux only)

randomstuff:
Just a proposition, in case someone is interested. Tested on x86 and x86_64 Linux with basic options.


skmp:
Hrfm, after some thinking, I'm still not quite ready to give up on the makefiles. Do you think it makes sense to have both in parallel on the repo?

randomstuff:
Yes, I guess it does:  the cmake may have some issues anyway. It adds some maintenance burden but it should not be too much.

skmp:
Okay, can you relocate the file to shell/cmake then pls ? :)

skmp:
Also, beware of https://github.com/reicast/reicast-emulator/issues/455

randomstuff:
Yes, OK. Moved and rebase agaainst master.

There is a strong expectation that the `CMakeLists.txt` is a the root of the source directory, though.

skmp:
That makes sense for cmake-based projects, but we have many build systems, and polluting the root dir doesn't look pretty to me. Looks good, I'm merging

skmp:
Gah, on a second thought can the toolchains be also moved into the cmake folder? *pretty pls*

randomstuff:
Yes, I forgot it. Sorry :)

randomstuff:
Updated.

PR #699, opened 2015-07-25
Build defines cleanup

skmp:
I've been hacking away on the rec-cpp branch, and I had to do some cleanups on the build defines to get everything working sanely. While I was at it I wrote down some ideas for it. What'ya think?

Figured I'd do this on a separate PR to get feedback (and spell checking...)


skmp:
@Holzhaus @LoungeKatt you might be interested ~

PR #701, opened 2015-07-25
Feat/rec cpp

skmp:
This is an idea for a dynarec that doesn't need to emit/jit compile code. 

Instead of compiling the blocks to assembly during runtime, I have a lot of "ready-to-be-configured" pre-compiled blocks. Then on dynarec time I pick the appropriate one, configure it, and let the normal dynarec/block manager process drive them.

The precompiled blocks are generated via templates, with block sizes from 1 to 512 il opcodes. I then have "decoding" matchers, that using stored register pointers will execute the canonical form of the opcodes. (Canonical is the name for the C implementation of the JIT IL).

The per-opcode idea is

```
void do_r32r32tor32_opcode(configured_op_context* op_ctx) {
    *op_ctx->rd = shil_add(*op_ctx->rs1, *op_ctx->rs2);
}
```

Ofc. there are many do_*_opcode functions, one for each type of parameters (r32, imms, floats, etc). The setup of the configured_op_context happens during jit time. 

On the block level it works like

````
void run_block(configured_block_context* block_ctx) {
    for (int i=0; i<OPCODE_COUNT; i++) {
        do_opcode_function_poiter* do_it = block_ctx[i]->op_function;
        do_it(block_ctx[i]->op_context);
    }
}
```

I'm naturally abusing templates and macros to generate the bulk of that stuff, so in the implementation there's a ton of meta programming and everything is wrapped in nested classes to make it work, etc, etc.

I have a ton more steps to get this to the idea I had, so it is likely to get some more big speed boosts. It already can run many games fulllspeed on my machine without any jit (msvc, x64, i7 @ 2.2ghz). I can dispatch 200 to 280 mops, which is amazing considering the interpreter barely gets to 30. 

If it sounds like fun, you might also enjoy reading about the ["The Nostradamus Distributor" by by Darek Mihocka](http://emulators.com/docs/nx25_nostradamus.htm) which presents many of the reasons why this is so fast, was a source of inspiration -- I read it many years ago and remixed the ideas in this impl. 

I also plan to write a blog post that goes into more detail and is in a more digestible format

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -37,6 +37,10 @@ ifdef X64_REC
     RZDCY_MODULES += rec-x64/
 endif
 
+ifdef CPP_REC
+    RZDCY_MODULES += rec-cpp/
+endif
+
 ifndef NO_REND
     RZDCY_MODULES += rend/gles/
 else
--- a/core/hw/pvr/spg.cpp
+++ b/core/hw/pvr/spg.cpp
@@ -86,6 +86,8 @@ extern double speed_load_mspdf;
 double speed_load_mspdf;
 #endif
 
+int mips_counter;
+
 double full_rps;
 
 u32 fskip=0;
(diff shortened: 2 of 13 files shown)


skmp:
Things are moving fast, so I'll be merging this one even though it looks ugly as F*CK. The good part is that the uglyness is isolated on rec-cpp so far. Emscripten and nacl work depend on this ~

Issue #702, opened 2015-07-26
[Linux] error: 'REG_E??' was not declared in this scope


Holzhaus:
I'm getting this when trying to compile:

```
../../core/linux/context.cpp: In function 'void context_segfault(rei_host_context_t*, void*, bool)':
../../core/linux/context.cpp:48:34: error: 'REG_EIP' was not declared in this scope
   bicopy(reictx->pc, MCTX(.gregs[REG_EIP]), to_segfault);
                                  ^
../../core/linux/context.cpp:17:62: note: in definition of macro 'MCTX'
 #define MCTX(p) (((ucontext_t *)(segfault_ctx))->uc_mcontext p)
                                                              ^
../../core/linux/context.cpp:49:35: error: 'REG_ESP' was not declared in this scope
   bicopy(reictx->esp, MCTX(.gregs[REG_ESP]), to_segfault);
                                   ^
../../core/linux/context.cpp:17:62: note: in definition of macro 'MCTX'
 #define MCTX(p) (((ucontext_t *)(segfault_ctx))->uc_mcontext p)
                                                              ^
../../core/linux/context.cpp:50:35: error: 'REG_EAX' was not declared in this scope
   bicopy(reictx->eax, MCTX(.gregs[REG_EAX]), to_segfault);
                                   ^
../../core/linux/context.cpp:17:62: note: in definition of macro 'MCTX'
 #define MCTX(p) (((ucontext_t *)(segfault_ctx))->uc_mcontext p)
                                                              ^
../../core/linux/context.cpp:51:35: error: 'REG_ECX' was not declared in this scope
   bicopy(reictx->ecx, MCTX(.gregs[REG_ECX]), to_segfault);
                                   ^
../../core/linux/context.cpp:17:62: note: in definition of macro 'MCTX'
 #define MCTX(p) (((ucontext_t *)(segfault_ctx))->uc_mcontext p)
```

skmp:
Bitness? Compiler? Linux distro?

Holzhaus:
I'm using ArchLinux:

```
$ gcc --version
gcc (GCC) 5.2.0
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ uname -a     
[...]

PR #703, opened 2015-07-26
Unified Linux makefiles (x86/x64)

Holzhaus:
This PR merges the lin86/lin64 Makefiles into one Makefile.

Buildarch is auto-detected via `uname -m`, but can also be enforced (which might be useful for cross-compiling):

```
make ARCH=i686
```

Compilation should work as soon as #702 has been fixed (that error occurs also with old makefiles after you replace `HOST_NO_AREC` with `TARGET_NO_AREC`).


skmp:
Hmm, can you make it print a warning and ask yes/no if x64 build is done? x86 build is more than 3 times faster right now.

skmp:
Also, maybe a linux-intel folder makes more sense? Since this doesn't handle arm and it's impossible variations

Holzhaus:
> Also, maybe a linux-intel folder makes more sense? Since this doesn't handle arm and it's impossible variations

Currently yes. But I'd like to merge in the Rapi2, beagle, etc stuff, too (at some point at least).

Holzhaus:
OK, this works now. PR #705 also merges the other Linux makefiles.

shell/linux/Makefile
```diff
@@ -2,18 +2,41 @@
...
 
-RZDCY_SRC_DIR = ../../core
+MFLAGS := 
+ASFLAGS := 
+LDFLAGS :=
+CFLAGS := 
+CXXFLAGS :=
 
-include $(RZDCY_SRC_DIR)/core.mk
+ifeq (,$(ARCH))
```
skmp: Wouldnt ?= also work?

PR #704, opened 2015-07-27
XMLParser and GLES 3 cleanup

LoungeKatt:


skmp:
The core doesn't quite support GLES 3 yet, but I guess it doesn't hurt using the context. So the covers are saved/cached now?

Overall looks good (but I didn't look into the details too much)

LoungeKatt:
It's not used, but it was there for the future. This is to avoid having to go back and fix it later.

Yes, the covers will be saved to the reicast external folder. Once the covers load the first time, the average time to load about 10 games with details and covers is just under 3 seconds.

PR #705, opened 2015-07-27
Merge Makefiles for beagle/pandora/gcwz/rapi2/lincpp into Linux Makefile

Holzhaus:
This is a follow-up PR to #703.

Hardware auto-detection code has been separated and can be overridden by using:

```
make platform=x86
make platform=rpi2
make platform=lincpp
# etc
```

**DISCLAIMER**: Since I do not own these devices, someone needs to test this before merging.


carstene1ns:
Just a short remark: You will likely never build directly on devices like the GCW-Zero, but instead cross-compile. So your platform detection will not work at all for those devices, but will detect the arch of the build machine. However, it should not matter much as you can specify a platform manually.

Holzhaus:
Yes, but as you already said it's possible to disable auto-detection by specifing the platform directly via `make platform=gcwz`.

The [platform auto-detection code](https://github.com/Holzhaus/reicast-emulator/blob/unified-makefiles-all/shell/linux/Makefile#L20-L45) is quite simple and short, so we should just leave it in for other devices. If we removed the gcwz part, that would save 2 lines :D

Furthermore, the GCW Zero might be a special case, but I'd appreciate something on almost every other platform, like generic x86/AMD64 Linux, Raspberry Pi 2, BeagleBone Black and ODROID-C1 (which I plan to add soon). I'm lazy and don't want want to setup a cross compilation environment if I can just SSH to that device, run `make` in a `screen` session and come back some hours (or even a day later) and everything is ready to use.

shell/linux/Makefile
```diff
@@ -0,0 +1,212 @@
...
+else ifneq (,$(findstring beagle,$(platform)))
+    CC_PREFIX ?= arm-none-linux-gnueabi-
+    MFLAGS += -marm -march=armv7-a -mtune=cortex-a9 -mfpu=neon -mfloat-abi=softfp -funroll-loops
+    ASFLAGS := -march=armv7-a -mfpu=neon -mfloat-abi=softfp
+    CFLAGS += -D TARGET_BEAGLE
+    USE_GLES := 1
+
+# Pandora
+else ifneq (,$(findstring pandora,$(platform)))
+    FOR_ARM := 1
```
Holzhaus: I guess I can remove this? `FOR_ARM` doesn't seem to be used anywhere except in the Pandora Makefile:
https://github.com/reicast/reicast-emulator/search?q=FOR_ARM

ptitSeb: Yeah, it's a left over from early version of the Makefile.

shell/linux/Makefile
```diff
@@ -0,0 +1,212 @@
...
+    CXXFLAGS += -fexceptions
+
+# Raspberry Pi 2
+else ifneq (,$(findstring rpi2,$(platform)))
+    MFLAGS += -marm -march=armv7-a -mtune=cortex-a9 -mfpu=neon -mfloat-abi=hard -funroll-loops
+    ASFLAGS += -march=armv7-a -mfpu=neon -mfloat-abi=softfp
+    CFLAGS += -D TARGET_BEAGLE -D TARGET_LINUX_ARMELv7 -DARM_HARDFP
+    INCS += -I/opt/vc/include/ -I/opt/vc/include/interface/vmcs_host/linux -I/opt/vc/include/interface/vcos/pthreads -I../linux-deps/include
+    LIBS += -L/opt/vc/lib/  -L../linux-deps/lib -lbcm_host
+    USE_GLES := 1
```
Holzhaus: Seems sane to me to change [Line 94 in `core/core.mk`](https://github.com/reicast/reicast-emulator/blob/39ead6159b69244ccd1be1796418990fe640fac1/core/core.mk#L94) from `ifndef DESKTOPGL` to `ifdef USE_GLES` to match this notation. Any objections?

shell/linux/Makefile
```diff
@@ -0,0 +1,212 @@
...
+# Generic x86_64 (64 bit)
+else ifneq (,$(findstring AMD64,$(platform)))
+    X64_REC := 1
+    NOT_ARM := 1
+    CFLAGS += -D TARGET_LINUX_x86 -D TARGET_NO_AREC
+    CXXFLAGS += -fexceptions
+
+# Raspberry Pi 2
+else ifneq (,$(findstring rpi2,$(platform)))
+    AS=${CC_PREFIX}gcc
```
Holzhaus: Just wondering: Is this correct? If I look at the rapi2 Makefile, it seems to use `gcc` instead of `as`.

Holzhaus: Hmm, seems to be neccessary, since it won't recognize the Macros and complain about "junk at the end of the line".

skmp:
yes, gcc should be used always. While working on NACL, I harmonized the NOT_ARM thingy (which makes no sense, really). Take a look at https://github.com/reicast/reicast-emulator/commit/d001e5f44c4631a0cb991ca685d07ecfcec73ee2

shell/linux/Makefile
skmp: Yeah, makes sense

skmp:
Also, detecting the hardware might not be enough -- a common pain w/ arm hardware is softfp abi vs hardfp.

Holzhaus:
We could detect hardfloat/softfloat via `gcc -dumpmachine`. But if a device has a floating point unit, most people will use that. All others are probably competent enough to edit the makefiles before building. At some point, we could try to build a more generic platform detection (don't check "Is this a Raspberry Pi 2?", but check "Does this device support neon/hardfloat/vfpv3/etc?", "is ALSA/X11/PulseAudio installed?"), but that could soon get very complex.

Since the `NOT_ARM` change is not merged yet, I'll wait until you merge that before I change this here.

About the ASM compiler: So if I understand correctly, I should change `$(AS)` to `gcc` for *all* platforms? Because in the current Makefiles, only the RPi2 is using `gcc`, all others are using `as`.

skmp:
hardfp or softfp is a calling rule (-mfloat-abi=XXX), they both use the hardware fpu. (-mfpu=soft doesn't). I know, confusing. Many distros insisted on softfp for binary compatibility w/ armv6, before eventually giving up and doing hardfp. Android is softfp too. Also, pandora afaik.

shell/linux/Makefile
```diff
@@ -11,47 +9,134 @@ USE_OSS := 1
...
+    NOT_ARM := 1
+    CFLAGS += -D TARGET_LINUX_x64 -D TARGET_NO_AREC -fsingle-precision-constant
     CXXFLAGS += -fexceptions
+
+# LinCPP
+else ifneq (,$(findstring lincpp,$(platform)))
+    CPP_REC := 1
+    NOT_ARM := 1
+    CFLAGS += -D TARGET_LINUX_x64 -D TARGET_NO_JIT
+    CXXFLAGS += -fexceptions -std=gnu++11
```
Holzhaus: Do the other implementation  support `-std=gnu++11` as well? If so, I could apply it globally, not just for LinCPP.

skmp:
There's no reason to compile everything w/ gnu++, it really limits the compatibility with older compilers. Only rec-cpp depends on it. It also breaks single precision consts.

Holzhaus:
Did anyone test this on beagle/pandora/gcwz/rapi2 yet?

skmp:
I'll test it on my rapi2 now that I have it back running!

skmp:
Building works, but x11 init fails on rapi. Debugging that

skmp:
Okay, rapi doesn't support -D SUPPORT_X11, but with that taken out it works nicely. There was some merge-weirdness that I had to fix -- can you rebase to current master and remove X11 from rpi2? Then feel free to merge (the other platforms aren't as major).

I've also added ya to the project, you can use nick/whatever for "private" branches, "feat/whatever" for "stable features", "wip/whatever" for "unstable-ideas-features" and "fix/whatever" for fixes!

(It makes my life easier not having to merge from different repos all the time, and keeps dev centralized on a single place)

Holzhaus:
OK. Strange thing: I had no rebase issues at all (no merge conflics whatsoever).

skmp:
Does it build though? :P

On Tue, Aug 11, 2015 at 5:38 PM, Jan Holthuis <notifications@github.com>
wrote:

> OK. Strange thing: I had no rebase issues at all (no merge conflics
> whatsoever).
>
> —
> <https://github.com/reicast/reicast-emulator/pull/705#issuecomment-129933619>
> .
>



-- 
~skmp

Holzhaus:
Didn't test that. I'll test on Linux_x64.

I just had are staring contest with the different makefiles and found that `-DSUPPORT_X11` is only enabled for `x86`, `x64`, `linCPP` and `pandora`. Thus I disabled `USE_X11` for `beagle` and `gcwz`, too. OK?

skmp:
Oh, look like master is broken. Fixing that, then it should be good to merge !

skmp:
Yeah. bealge is kind of dead anyways, and gcwz is unpractical as things (but good to have anyway).

Holzhaus:
> Oh, look like master is broken
Yeah, commit 45f0a47eec3f7a7d16cadbb1d16d0a2f5e46c78c broke `core/linux-dist/main.cpp`:

```
g++   -I../../core -I../../core/deps -I../../core/khronos  -D TARGET_LINUX_x64 -D TARGET_NO_AREC -fsingle-precision-constant -g -O3 -D RELEASE -c -D USES_HOMEDIR  -frename-registers -fno-strict-aliasing -ffast-math -ftree-vectorize -fomit-frame-pointer -D SUPPORT_X11   -fexceptions -fno-rtti -fpermissive -fno-operator-names -D SUPPORT_X11  -D USE_ALSA -D USE_OSS ../../core/linux-dist/main.cpp -o obj-x64/linux-dist/main.build_obj
../../core/linux-dist/main.cpp:451:1: error: stray '\302' in program
  
 ^
../../core/linux-dist/main.cpp:451:1: error: stray '\240' in program
../../core/linux-dist/main.cpp:4:0: error: unterminated #if
 #if HOST_OS==OS_LINUX
 ^
Makefile:210: recipe for target 'obj-x64/linux-dist/main.build_obj' failed
make: *** [obj-x64/linux-dist/main.build_obj] Error 1
```

LoungeKatt:
Looks like the closing endif got removed.

LoungeKatt:
Fixed.

skmp:
Lol :p. That entire code needs a nice rewrite one of these days...

PR #714, opened 2015-07-29
Wip/nacl and emscripten

skmp:
Fully rebased, and mostly merged between emscripten and nacl.

~~I still have to check emscripten as the sdk is broken, and a reinstall didn't fix it (on windows, linux or osx)~~ *Mozilla's installer worked, emscripten is now fixed*


skmp:
Android builds and runs, nacl builds and runs, emscripten builds and runs. linux 86 and x64 build and run. I'm in huge favor of merging this :D

skmp:
The only thing I see broken is that FEAT_HAS_COREIO_HTTP isn't correctly set (but practically no-one uses this feature atm, cus it is uber slow w/o async & cache)

PR #718, opened 2015-07-29
hw/sh4/dyna/driver.cpp: Fix pointer size on 64 bit arch

Holzhaus:
On AMD64/x86_64 systems, `sizeof(int)` is 4 bytes but `sizeof(int*)` is 8 bytes.

If you try to compile this with `clang`, you'll get:

```
../../core/hw/sh4/dyna/driver.cpp:311:9: error: cast from pointer to smaller type 'int' loses information
        verify((int)rbi);
               ^~~~~~~~
../../core/types.h:521:23: note: expanded from macro 'verify'
#define verify(x) if((x)==false){ msgboxf("Verify Failed  : " #x "\n in %s -> %s : %d \n",MBX_ICONERROR,(__FUNCTION__),(__FILE__),__LINE__); dbgbreak;}
```

### DIFF
--- a/core/hw/sh4/dyna/driver.cpp
+++ b/core/hw/sh4/dyna/driver.cpp
@@ -309,7 +309,7 @@ void* DYNACALL rdv_LinkBlock(u8* code,u32 dpc)
 		rbi=bm_GetStaleBlock(code);
 	}
 	
-	verify((int)rbi);
+	verify((size_t)rbi);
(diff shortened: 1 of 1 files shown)


Holzhaus:
Looks like this has been fixed by commit 8c68f4a06bea64585176d0679b436ee9e059ad40. Closing this.

PR #721, opened 2015-08-02
Convert third-party source to library (avoid license conflict)

LoungeKatt:


LoungeKatt:
This is somewhat sloppy at the moment, but it gets all the stuff that can't be relicensed into a folder that can be excluded from the blanket license. This will be followed by actually integrating the libraries in their original condition, rather than the current bulk library project.

Holzhaus:
If these third-party libraries have a git repo, maybe we can use them via [`git submodule`](https://git-scm.com/book/en/v2/Git-Tools-Submodules) instead of hosting the code ourselves?

LoungeKatt:
That would work, had it not been for the customization to them not included in the upstream. There is also the instability created by connecting a submodule to a repo not maintained by this project. That is just asking to invite error.

Holzhaus:
Submodules are locked at a specific commit, so if the upstream repo is updated, nothing will change for us until we update the submodule. The only instability I can think of is when the upstream repo is taken offline.

Have a look at: http://stackoverflow.com/questions/1979167/git-submodule-update/1979194#1979194

LoungeKatt:
I'm aware of how they work, but did not use it INTENTIONALLY.
Submodule does not account for modifications not included in the upstream.

LoungeKatt:
For reference sake, the original repository for SwitchCompat is https://github.com/ankri/SwitchCompatLibrary which is 3 years old and does not include Preferences. I overhauled the library, but must retain the original license since it is derivative.

Holzhaus:
> I'm aware of how they work, but did not use it intentionally.
Submodule does not account for modifications not included in the upstream.

Yes, i was referring to the instability concerns you raised. Didn't realize the original project is unmaintained, otherwise I'd suggest to try to push these changes upstream.

Anyway, good that someone takes care of these license issues. :+1:

LoungeKatt:
The authors were not interested in merging the code. It was submitted years ago when the repositories were still active. Since then, 9oldandroids and other "full-service" libraries have replaced them but are overkill and a little too "needy" for things like checkboxes and a single menu.

skmp:
Looks okay to me (wth is iCade though?). @Holzhaus submodules only ever bring pain, that's the sad truth...

LoungeKatt:
@skmp It will support wireless controllers for the iPhone version. At least, it should once it is finished. I am still updating the code.

http://www.ionaudio.com/products/details/icade
http://www.gametel.se/
http://8bitdo.com/nes30pro/

skmp:
Sweet :D

PR #724, opened 2015-08-03
Fix some clang warnings and errors

### LINKED ISSUE #719: Fix clang compilation errors
Trying to compile with `clang` yielded some interesting bugs. I was able to fix #718, but then I got this one which I'm not able to fix (because I don't understand the code well enough).

http://pastebin.com/H7K92vEw

Holzhaus:
Resolves #719 and some warnings (mostly format strings and missing parentheses).

### DIFF
--- a/core/hw/holly/sb_mem.cpp
+++ b/core/hw/holly/sb_mem.cpp
@@ -216,7 +216,7 @@ void  DYNACALL WriteMem_area0(u32 addr,T data)
 	const u32 base=(addr>>16);
 
 	//map 0x0000 to 0x001F
-	if ((base >=0x0000) && (base <=0x001F) /*&& (addr<=0x001FFFFF)*/)// :MPX System/Boot ROM
+	if ((base <=0x001F) /*&& (addr<=0x001FFFFF)*/)// :MPX System/Boot ROM
 	{
 		//EMUERROR4("Write to  [MPX	System/Boot ROM] is not possible, addr=%x,data=%x,size=%d",addr,data,sz);
 		WriteBios(addr,data,sz);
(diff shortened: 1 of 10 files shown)


Holzhaus:
Unsure if we want to use `%z` instead of `%lu`to print a `size_t`? Any opinions?

skmp:
After a bit of reading around, %lu seems to be not working always correctly (long is 4 bytes on x64 windows). I compiled and it runs O.K. But I gotta check for it...

skmp:
Yes, after testing, `printf("%lu %d\n", (size_t)0xFFFFAFFFFFFFFFFE, 3)` fails on windows, and needs `%Iu`

Holzhaus:
Just saw that `%zu` doesnt work on windows either. We either need to dereference the pointer and use `%p` (ugly) or use  use `%I64x` (also ugly) or do some [macro hacking](http://stackoverflow.com/a/1669254/1455128) (even more ugly).

randomstuff:
About the `call((void*) foo)` issue, I submitted a [feature request](https://github.com/herumi/xbyak/issues/23) to xbyak because this was `call(foo)` was causing warning on GCC. With variadic template support (C++11), it is now possible to have a standard compliant `call(foo)` with recent versions of xbyak.

However enabling C++11 breaks the compilation of reicast (at least on my computer) because some standard headers are now pulled and are not happy about the `-fsingle-precision-constant` flag.

skmp:
Οκαυ, after running over this once more can you keep the non-formatting fixes? The formatting changes break windows, which is the primary development platform atm.  @randomstuff thx, I'll update our local xbyak copy :dancer:

Holzhaus:
I already cherry-picked and merged these fixes in #753. ;-)

skmp:
Ahh I missed that. The void* casts shouldn't be needed after #804 is merged, but they don't hurt either. :+1:

PR #727, opened 2015-08-05
rend/soft: Import the experimental core from nulldc/drkpvr, better re…

skmp:
…nderer modularization

- Adapted for indexed, async rendering, shared ta decoder
- This blits via a quick-and-hacky GDI blitter for now
- SSE/MMX intrins based, so not very portable
- A rather not good "reference" implementation
- At least, it's not terribly slow though
- GetTexture moved to Renderer interface

### DIFF
--- a/core/hw/pvr/Renderer_if.cpp
+++ b/core/hw/pvr/Renderer_if.cpp
@@ -348,7 +348,20 @@ bool rend_init()
 #else
 
 #if HOST_OS == OS_WINDOWS
-	renderer = settings.pvr.rend == 0 ? rend_GLES2() : rend_D3D11();
+	switch (settings.pvr.rend) {
+		default:
+		case 0:
+			renderer = rend_GLES2();
+			break;
+
+		case 1:
+			renderer = rend_D3D11();
+			break;
+
+		case 2:
+			renderer = rend_softrend();
+			break;
+	}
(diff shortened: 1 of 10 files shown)


skmp:
The code is from some attempt for a quad-based renderer I co-designed with @gigaherz like, in 2007? or dunno. I don't remember the details, but the attempt was to be a simd-pixel-parallel implementation, and I got meh'd on computing the texcoords/textdata in parallel. Runs fairly fast though, even with GDI blitting.

Bios menu

Soul calibur ingame

LoungeKatt:
Windows only or would it make sense to connect the Android option for software rendering to this instead ? Especially considering using actual Android software rendering is like taking a steamer on the emu.

skmp:
This is clearly x86 only, windows only atm. I want to make 2 softrend versions, one aiming for 100% accuracy and one aiming for portable speed, which I'd like to include on android. But something is better than nothing.

skmp:
Also, interestingly, x64 is almost twice as fast on rendering as x86 (on msvc release build)

LoungeKatt:
I figured as much by the one HOST_OS that showed up on the changes. Even so, I have a conversion for the Android pvr GUI option whenever the time comes. I was poking around in there anyway.

PR #728, opened 2015-08-05
Feat/x64 all blockends

skmp:


skmp:
And with this, x64 is almost as fast as x86. To be fair, x64 doesn't include DSP, and doesn't run via VirtARM (which is 2x slower)

PR #729, opened 2015-08-05
Wip/softrend

skmp:
Hacking around w/ some ancient nullDC softrend code I had on my disk. I'm quite certain the triangle math is broken/doesn't completely make sense, however it works well enough to be used for testing and gets respectable (20-30) fps even w/ texturing on my old lappy.

Screens ~
ct-ingame-softrend
doa2-demo-softrend
sc-ingame-softrend
sotb-menu-softrend
sotb-softrend

Meanwhile, during dev it was more like
jgr-psytext-1
jgr-psytext-2

### DIFF
--- a/core/hw/pvr/pvr_regs.h
+++ b/core/hw/pvr/pvr_regs.h
@@ -419,7 +419,7 @@ union FPU_SHAD_SCALE_type
 
 
 #define FPU_SHAD_SCALE    PvrReg(FPU_SHAD_SCALE_addr,FPU_SHAD_SCALE_type) // RW  Intensity Volume mode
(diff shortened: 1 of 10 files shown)


LoungeKatt:
@skmp I rebased this just to make sure the last commit failing wasn't the recent merge

skmp:
_mm_free & _mm_malloc are the issue on texcache. I either need to `#ifdef` them or do a refactor of the  texcache code. The refactor is meh work, mostly because the rest of the code is so ugly, but I guess that's the way to go... Then we can merge to mainline!

skmp:
Decided having a single source is worth more than the cleanups, which would make me postpone the merge and lead to merge conflicts. Let's see if it builds so that we merge it!

LoungeKatt:
Haha, nice!

PR #731, opened 2015-08-08
Rename newdc to reicast

skmp:


skmp:
@LoungeKatt this won't break any of the android log stuff, right?

LoungeKatt:
I can easily update it. I'll put a pull in shortly.

PR #743, opened 2015-08-11
Feat/nvmem automatic fallback

skmp:
This makes #652 obsolete by implementing automatic fallback. Needs to be tested with problematic hardware, and we're pretty much set for r7 from a technical PoV


skmp:
Rebased to master

PR #648, opened 2015-05-08
Feat/less video stuttering

skmp:
Trying to guess what causes #602

PR #744, opened 2015-08-11
sync: Also sync to video if too fast

skmp:
This replaces #648 in a nicer way. Needs to be tested with problematic hardware.

### DIFF
--- a/core/hw/pvr/ta_ctx.cpp
+++ b/core/hw/pvr/ta_ctx.cpp
@@ -1,6 +1,8 @@
 #include "ta.h"
 #include "ta_ctx.h"
 
+#include "hw/sh4/sh4_sched.h"
+
 extern u32 fskip;
 extern u32 FrameCount;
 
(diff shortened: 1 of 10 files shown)


skmp:
@LoungeKatt we also need an option for this for android ~ *pretty pls*

LoungeKatt:
Gotcha. Should that override the Turbo stuff or do you think they'd play nice?

skmp:
Btw, dreamcast builds were broken after the naomi merge, just fixed on master...

skmp:
Turbo would conflict w/ this, as it caps the top speed to have more fluid visuals. In turbo mode, it should be disabled. We need to also overhaul how the settings stuff gets applied and all..

LoungeKatt:
It was still building Android. Got lucky I guess. 

It disables turbo. You can still turn sound off, though.

skmp:
I'm still not sure this is better than not having a fix at all, but it's a starting point for future work so I'll merge it in. ~~

PR #748, opened 2015-08-15
Fix path structure and relationships in iOS project

LoungeKatt:
@skmp owes me the blood of a newborn child or whatever it is for this one.

That was a PITA to figure out, but it seems you need a second tier under the project file to satisfy the "group" relation for everything else. Doing it relative to the project file is even more messy.


skmp:
:D :D :D :D :D :D :D

PR #745, opened 2015-08-12
Clean up linux-dist/main.cpp and make evdev/joystick device configurable

Holzhaus:
Wow, this one looks bigger than it actually is. This one mainly fixes identation (that's why this changes so many lines), parameter spacing, etc. I also unraveled the evdev/joystick code a bit.

PR #751, opened 2015-08-15
Fix linux-dist/main.cpp formatting

Holzhaus:
This are just some aesthetic changes (formatting, whitespace, etc.).

Replaces the cleanup part of #745.

### DIFF
--- a/core/linux-dist/main.cpp
+++ b/core/linux-dist/main.cpp
@@ -16,65 +16,64 @@
 #include <unistd.h>
 
 #if defined(TARGET_EMSCRIPTEN)
-	#include <emscripten.h>
+  #include <emscripten.h>
 #endif
 
-
-
 #if defined(SUPPORT_X11)
-	#include <X11/Xlib.h>
-	#include <X11/Xatom.h>
-	#include <X11/Xutil.h>
+  #include <X11/Xlib.h>
+  #include <X11/Xatom.h>
+  #include <X11/Xutil.h>
 
-	#if !defined(GLES)
-		#include <GL/gl.h>
-		#include <GL/glx.h>
-	#endif
-	
-	#include <map>
-	map<int, int> x11_keymap;
+  #if !defined(GLES)
+    #include <GL/gl.h>
+    #include <GL/glx.h>
+  #endif
+
+  #include <map>
+  map<int, int> x11_keymap;
 #endif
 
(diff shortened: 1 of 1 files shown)


LoungeKatt:
I was always told compiler ifdefs should not be indented. Was there a specific reason to indent them along with everything else?

Holzhaus:
Better readability. As far as i know, back in the pre-ANSI days, compilers expected lines containing a preprocessor directive to start with a `#` character (so you couldn't indent them, because the line would then start with a space or tab character).

In modern C, this is not true anymore. I guess people still avoided indenting `#ifdef` in case someone uses an ancient (pre-ANSI C) compiler version. Today, I don't think they exist anymore.

Anyway, since we're doing C++ instead of C, we don't need to bother with this, since it has always been like that.

> A preprocessing directive consists of a sequence of preprocessing tokens. The first token in the sequence is a# preprocessing token that is either the first character in the source file (optionally after white space containing no new-line characters) or that follows white space containing at least one new-line character. The last token in the sequence is the first new-line character that follows the first token in the sequence.
*Source: [ISO/IEC 14882:1999 (C++ standard), p. 302](http://tracker.ceph.com/attachments/download/752/cpp98.pdf)*

So IMHO there isn't really a reason not to indent them.

LoungeKatt:
Ironically, Unity automatically removes any indentation when you type a directive. It seems like a rather tedious task to manually format something for no other reason than it can be indented. I thought there might have been a benefit you knew that I was missing.

skmp:
At last got some time to look on the changes! After a bit on thinking on it, I'd rather we stick to tab-based formating (I use 1 tab = 4 spaces). It's the visual studio default and it's what I'm most used to work with. Is there some strong reason to go with space-formating for the linux side?

Holzhaus:
> Is there some strong reason to go with space-formating for the linux side?

Nope. I didn't even notice you we're using tabs, sorry :D

I personally prefer space, especially when you want to do something like:

```
if (foooo(somethingreallyreallyreallyreallyreallyreallylong1,
              somethingreallyreallyreallyreallyreallyreallylong2)
```

(you can't really align that with just tabs).

But that's not that important, so I can change it back to tabs (preferably after we merged #752, because I really don't want to end up in rebase hell).

skmp:
Yup, sounds like plan. I'll reboot and then test the changes and merge if all is good!

PR #752, opened 2015-08-15
Linux Input overhaul (Better evdev input & multiplayer support)

Holzhaus:
This is a follow up to PR #751 and replaces the remaining parts of PR #745.


zenogears:
Hi!
LT/RT on linux do not working (not at q/w or a/s on keyboard)
=\
How to fix this?

Holzhaus:
Don't know if we really want that last commit. It removes the x11 keyboard input in favour of evdev.

Pros:
- Left/Right trigger works now
- You can use the keyboard on non-X11 platforms

Cons:
- You can no longer use controller and keyboard at the same time
- You need to specify the evdev device id of the keyboard before you can use it

I think both Cons can be fixed eventually, but currently they are not.

skmp:
I think we can keep both X11 and evdev (and specify on the config files). I'm sure some platform has like x11 working but evdev broken .....

Holzhaus:
OK, you can now disable x11 keybard input via config. I also fixed a nasty bug with mapping autodection when using the `xboxdrv` driver.

Holzhaus:
\*sigh\*, I had to separate xpad and xboxdrv mapping, since xboxdrv seems to use different axis codes for the left/right triggers by default than xpad. Anyhow, now both should work fine out of the box.

Holzhaus:
Just a quick overview what this PR does:
- code refactored and separated
- both evdev and the joystick api can be deactivated separately at compile time
- evdev, joystick api & x11 keyboard can be deactivated separately at runtime (via config)
- you can now supply your own devices to evdev/joystick api via config file instead of hardcoding them
- enabled multiplayer support (4 players) with evdev
- Added controller Mappings for:
  - xbox360 controller (wired and wireless) with xpad driver
  - xbox360 controller with xboxdrv driver (default mapping)
  - generic controller
  - generic keyboard
  - plus the already existant pandora & gcwz mappings

**Usage:**

Check what you event devices are:

```
$ evtest
Available devices:
/dev/input/event0:      cec_input
/dev/input/event1:      AT Translated Set 2 keyboard
/dev/input/event2:      Microsoft X-Box 360 pad
/dev/input/event3:      Microsoft X-Box 360 pad
/dev/input/event4:      Microsoft X-Box 360 pad
/dev/input/event5:      Microsoft X-Box 360 pad
```

Then map them in you config and disable joystick/x11 input for now:

```ini
[input]
evdev_device_id_1 = 2     # Player 1 uses the first Xbox pad
evdev_device_id_2 = 1     # Player 2 uses the keyboard
evdev_device_id_3 = 3     # Player 3 uses the second Xbox pad
evdev_device_id_4 = -1    # Player 4 is disabled
enable_x11_keyboard = -1  # X11 keyboard support is disabled
joystick_device_id=-1     # legacy joystick api is disabled
```

Please test ;-)

randomstuff:
I"m gettings:

<pre>../../core/linux-dist/evdev_mappings.cpp:57:3: error: uninitialized const member ‘s_evdev_controller_mapping::Axis_DPad_Y’
../../core/linux-dist/evdev_mappings.cpp:57:3: error: uninitialized const member ‘s_evdev_controller_mapping::Axis_DPad2_X’
../../core/linux-dist/evdev_mappings.cpp:57:3: error: uninitialized const member ‘s_evdev_controller_mapping::Axis_DPad2_Y’
../../core/linux-dist/evdev_mappings.cpp:57:3: error: uninitialized const member ‘s_evdev_controller_mapping::Axis_Analog_X’
../../core/linux-dist/evdev_mappings.cpp:57:3: error: uninitialized const member ‘s_evdev_controller_mapping::Axis_Analog_Y’
../../core/linux-dist/evdev_mappings.cpp:57:3: error: uninitialized const member ‘s_evdev_controller_mapping::Axis_Trigger_Left’
../../core/linux-dist/evdev_mappings.cpp:57:3: error: uninitialized const member ‘s_evdev_controller_mapping::Axis_Trigger_Right’
</pre>

After adding 8 `-1` entries to `controller_mapping_keyboard`, it's compiling allright.

Then I can get somehow into the game but I can't get too far because my gamepad is not a Xbox one and the default configuration does not seem OK for it. I probably have to configure the suitable bindings for it.

At some point, reicast dies with:

<pre>Fatal error : death by escape key
 in input_evdev_handle -> ../../core/linux-dist/evdev.cpp : 104
</pre>

I'll try to investigate this later.

randomstuff:
It would be great in the (distant) future to be able to to specify the device with it's name and an index. This would fix the issue where [the evdev index changes](https://code.google.com/p/dolphin-emu/issues/detail?id=8781) when the device is unplugged/replugged or on reboot.

Holzhaus:
@randomstuff I fixed the compile error you were getting.

About the crappy mapping for generic controllers: Either add a new mapping to [evdev_mappings.cpp](https://github.com/reicast/reicast-emulator/blob/holzhaus/linux-input-overhaul/core/linux-dist/evdev_mappings.cpp) (you can find out the codes with `evtest` or use `xboxdrv` to [emulate the xbox 360 controller](http://steamcommunity.com/app/236090/discussions/0/558748653724279774/).

The `death by escape key` error comes up, because you pressed the key that is used to exit the emulator (the SELECT button on the Xbox 360 pad), so that's expected.

> It would be great in the (distant) future to be able to to specify the device with it's name and an index. This would fix the issue where the evdev index changes when the device is unplugged/replugged or on reboot.

Yup, good idea. I've even thought about some auto-detection algorithm to have a better out-of-the-box experience.

randomstuff:
I tested the branch and [added support for my gamepad](https://github.com/randomstuff/reicast-emulator/commit/33b96ec959a013b05538ddd4d1541c09193b1c74) (in the long term it'd be nice to be able to change the mapping in the configuration instead of doing it in the code) (Yes, I still refuse to use xboxdrv for this…).

It is currently refusing to use the analog stick however. It seems to be configured correctly.

Shoegzer:
@Holzhaus really nice stuff here, thanks for this.  

I have two Sony DS4 pads with me and can help test with them if needed (kernel 3.16 here).  I haven't tried your new source yet but I'd be happy to jump in if you think testing would help.  

I agree with randomstuff that it would be nice to change mapping in the config instead of code - but baby steps.  Speaking of which - I'm sure this is way ahead of things but it would eventually be nice to see [url=http://www.segaretro.org/Dreamcast_Jump_Pack]Jump Pack[/url] (rumble) support, and the Sega Keyboard for TOTD (though not sure if/how that would conflict with an actual PC keyboard with mapped keys as discussed earlier).

Holzhaus:
@randomstuff 
> It is currently refusing to use the analog stick however. It seems to be configured correctly.

Mind sharing the `Supported Events` section of the `evtest` output? I suspect the Max/Min values of that axis are different from the xbox360 controller.

> in the long term it'd be nice to be able to change the mapping in the configuration instead of doing it in the code

Guess what I'm currently working on ;-)

@Shoegzer 
> I have two Sony DS4 pads with me and can help test with them if needed (kernel 3.16 here). I haven't tried your new source yet but I'd be happy to jump in if you think testing would help.

Testing is always good. Please try to emulate the 360 controller with `xboxdrv` by the way. I want to make sure that at least the 360 controllers work, so that we can merge this (xbox360 controller is better than broken input for *all* controllers ;-))

If we've merged this, I'll make a new PR that improves things and adds mapping file support.

randomstuff:
OK, I always managed to avoid using xboxdrv but I'll try to make it work in order to test this.

randomstuff:
This is the output of evtest:

<pre>Input device ID: bus 0x3 vendor 0x44f product 0xb323 version 0x100
Input device name: "Thrustmaster Dual Trigger 3-in-1"
Supported events:
  Event type 0 (EV_SYN)
  Event type 1 (EV_KEY)
    Event code 304 (BTN_SOUTH)
    Event code 305 (BTN_EAST)
    Event code 306 (BTN_C)
    Event code 307 (BTN_NORTH)
    Event code 308 (BTN_WEST)
    Event code 309 (BTN_Z)
    Event code 310 (BTN_TL)
    Event code 311 (BTN_TR)
    Event code 312 (BTN_TL2)
    Event code 313 (BTN_TR2)
    Event code 314 (BTN_SELECT)
    Event code 315 (BTN_START)
  Event type 3 (EV_ABS)
    Event code 0 (ABS_X)
      Value      0
      Min        0
      Max      255
      Flat      15
    Event code 1 (ABS_Y)
      Value      0
      Min        0
      Max      255
      Flat      15
    Event code 2 (ABS_Z)
      Value      0
      Min        0
      Max      255
      Flat      15
    Event code 3 (ABS_RX)
      Value      0
      Min        0
      Max      255
      Flat      15
    Event code 4 (ABS_RY)
      Value      0
      Min        0
      Max      255
      Flat      15
    Event code 5 (ABS_RZ)
      Value      0
      Min        0
      Max      255
      Flat      15
    Event code 16 (ABS_HAT0X)
      Value      0
      Min       -1
      Max        1
    Event code 17 (ABS_HAT0Y)
      Value      0
      Min       -1
      Max        1
  Event type 4 (EV_MSC)
    Event code 4 (MSC_SCAN)
  Event type 21 (EV_FF)
    Event code 80 (FF_RUMBLE)
    Event code 81 (FF_PERIODIC)
    Event code 88 (FF_SQUARE)
    Event code 89 (FF_TRIANGLE)
    Event code 90 (FF_SINE)
    Event code 96 (FF_GAIN)
</pre>

Holzhaus:
@randomstuff 
There's the problem. The analog stick axes of the xbox360 controller range from `-32768` to `+32767`, and because the Dreamcast's range is `-128` to `+127`, the values are [divided by 256](https://github.com/reicast/reicast-emulator/blob/holzhaus/linux-input-overhaul/core/linux-dist/evdev.cpp#L192-L196). The axes of your controller has a range from `0` to `255`, so their values will always be 0 (integer division).

randomstuff:
With xboxdrv, it's working OK (including the analog stick):

~~~sh
xboxdrv --evdev /dev/input/event17 --evdev-absmap ABS_X=x1,ABS_Y=y1,ABS_Z=x2,ABS_RZ=y2,ABS_HAT0X=dpad_x,ABS_HAT0Y=dpad_y --evdev-keymap BTN_B=a,BTN_C=b,BTN_A=x,BTN_X=y,BTN_Y=lb,BTN_Z=rb,BTN_TL=lt,BTN_TR=rt,BTN_TR2=start,BTN_TL2=select,BTN_SELECT=tl,BTN_START=tr
~~~

Shoegzer:
@Holzhaus sounds good.  I assume the ultimate goal is to "future-proof" input support so that it would work with any future controller no matter the make/model?  If that's possible, awesome.  :)

randomstuff:
I tested the branch some more with xboxdrv and did not find any issue.

randomstuff:
Indeed patching with `joyy[port] = ie.value - 128;` fixes the analog stick. But just for me :)

zenogears:
patch for [ShanWan Twin USB Joystick](https://github.com/zenogears/reicast-emulator/commit/1e665e07ac8bec21850c650e626ee7da20cc0743), but it's working weird...
Buttons worked as pressed and unpressed (<1ms), so character moving like... little step, one more, another..
i think somewhere is mistake...

randomstuff:
My gamepad has two analog triggers but they are somewhat reverted: 255 if not-pressed and 0 is fully-pressed so I had to patch the `Axis_Trigger_Left` and `Axis_Trigger_Right` with `lt[port] = 255 - ie.value;` as well.

Holzhaus:
@skmp Looks like this has been tested quite a bit. Can we merge?

Holzhaus:
@zenogears You have patched the master branch, not this one. Also, you were using the legacy joystick api (which doesn't work correctly for me either) instead of the evdev api.

zenogears:
@Holzhaus Yap. Your brunch after compilation doing [segmentation fault (core dumped)](http://pastebin.com/Rd7cRAqQ)

skmp:
@Holzhaus iyeah. I'm reading the diff and merging unless I find smth off :+1:

skmp:
Only objection is initializing a full 4-controller set by default. Can you make this a config? Otherwise, it looks good.

I have to refactor the shared stuff on my side later. This is sooo muchh nicer than the older code <3

Holzhaus:
Yeah. The main problem is, that the config file is loaded within dc_init(). We need to move that out of there, so we can read the config *before* initializing the emulator, but that'd mean we need to change the code of all other platforms, too.

skmp:
So mcfg init is before config load? That sounds wrong :p. Can you make an
#ifdef for linux then?

In most platform the input doesn't check the port number, and this would
create 4 controllers with mirrored inputs

On Tue, Aug 18, 2015 at 3:23 PM, Jan Holthuis <notifications@github.com>
wrote:

> Yeah. The main problem is, that the config file is loaded within
> dc_init(). We need to move that out of there, so we can read the config
> *before* initializing the emulator, but that'd mean we need to change the
> code of all other platforms, too.
>
> —
> <https://github.com/reicast/reicast-emulator/pull/752#issuecomment-132203151>
> .
>



-- 
~skmp

Holzhaus:
Well, both [`cfgOpen()`](https://github.com/reicast/reicast-emulator/blob/71c2cb4512d91f14fd22383514548391c9fa64b1/core/nullDC.cpp#L162) and [`mcfg_CreateDevices()`](https://github.com/reicast/reicast-emulator/blob/71c2cb4512d91f14fd22383514548391c9fa64b1/core/nullDC.cpp#L211) are called inside `dc_init()`.

Right now, I'm doing this in the linux `main.cpp`:

```c++
dc_init();
//now read config values & set everything up
dc_run();
```

This doesn't work if I want to enable/disable stuff after calling `dc_init()` obviously ;-)

So I'd propose to move the `cfgOpen()` call out of `dc_init()` and into `main.cpp` (and the files for the other platforms, too), so that we do this:

```c++
cfgOpen();
//now read config values & set everything up
dc_init();
dc_run();
```

skmp:
Nah, I don't want to move reicast-specific code to platform-specific code.
Just looked at the code and it is sane enough. The contract is "all init
happens after cfg file, apart from nvmem and command line". nvmem needs to
be the first thing done before memory fragmentation, command line so that
you can select an alternative cfg file.

Holzhaus:
Then maybe move `cfgOpen()` into a new function like `dc_preinit()`? :wink:

The point is: right now I need to call `dc_init()` before I can read config values. But when `dc_init()` returns, it's already too late to enable/disable controllers, because `mcfg_CreateDevices()` has already been called.

Holzhaus:
@randomstuff @Shoegzer 
Psst..hey you kids... Wanna use custom mapping files? Check out #769. (Axis range detection not done yet).

skmp:
You can just use the config on mcfg_CreateDevice. nullDC used to do it this
way (as part of the plugin-init). You can keep a map for all of the
controllers -- like [maple] controller_%d = whatever config here. I'll be
on irc in an hour or so if you want to discuss !

On Tue, Aug 18, 2015 at 10:02 PM, Jan Holthuis <notifications@github.com>
wrote:

> @randomstuff <https://github.com/randomstuff> @Shoegzer
> <https://github.com/Shoegzer>
> Psst..hey you kids... Wanna use custom mapping files? Check out #769
> <https://github.com/reicast/reicast-emulator/pull/769>. (Axis range
> detection not done yet).
>
> —
> <https://github.com/reicast/reicast-emulator/pull/752#issuecomment-132333692>
> .
>



-- 
~skmp

Holzhaus:
I'd rather let the host platform decide, i.e. if I specified valid controllers in the `[input]` section and they are connected - use 4 controllers. If I added a `[maple]` section, we also need to enable them there (so we're essentially doing the same thing twice).

Anyhow, I feel like this is something for another PR. I'll undo that specific commit so that we can merge and put it into another branch.

Holzhaus:
> Only objection is initializing a full 4-controller set by default. Can you make this a config? It could break input for other badly-coded platforms. Otherwise, it looks good.

I removed that commit for now. Can I merge?

skmp:
Yes!

On Wed, Aug 19, 2015 at 2:08 PM, Jan Holthuis <notifications@github.com>
wrote:

> Only objection is initializing a full 4-controller set by default. Can you
> make this a config? It could break input for other badly-coded platforms.
> Otherwise, it looks good.
>
> I removed that commit for now. Can I merge?
>
> —
> <https://github.com/reicast/reicast-emulator/pull/752#issuecomment-132565157>
> .
>



-- 
~skmp

skmp:
:+1: :+1:

Holzhaus:
;-)

PR #755, opened 2015-08-16
cfg: Make ConfigFile class reusable

Holzhaus:
- Moved the code for accessing config entries and parsing config files into the actual `ConfigFile` class and decoupled it from the `cfgdb` variable.
- Put the `ConfigFile` declaration into the header file, so that we can reuse it somewhere else.
- Added the possibility to load hex strings (e.g. `0xAF`) via `ConfigFile::LoadInt()`
- Removed some redundant code

I thought this might be useful anyway, but I also did this because I'm planning to dump the [hardcoded evdev mappings](https://github.com/reicast/reicast-emulator/blob/holzhaus/linux-input-overhaul/core/linux-dist/evdev_mappings.cpp) and read them from a mapping file instead. This way, you could also use a completely custom mapping.


skmp:
It would also be great if you can move the cfg class outside the cfg.h header, as It's not really needed to be exposed to the rest of the emulator. Perhaps cfg/ini.cpp && cfg/ini.h and then have just the Load/Store functions on cfg.cpp/h? Does that sound reasonable?

Holzhaus:
Done.

skmp:
I guess next up the cfg list is to parse w/ C++ streams for buffer overflows. /me strongly dislikes C++ streams

Holzhaus:
That code is somehow buggy anyway. I can't loop over it for some reason (always segfaults). Or maybe I'm just stupid.

Maybe just reimplement it using `std::map`?

skmp:
Yeah, that wouldn't be bad either.

PR #757, opened 2015-08-17
Fix Linux Makefile

randomstuff:
`as` does not recognize `-m32`


Holzhaus:
This breaks ARM support. Check out #762.

PR #760, opened 2015-08-17
Update project for #755 and fix warnings

LoungeKatt:
@skmp can you verify fixing these warnings doesn't break anything for you?

Tested Android and iOS and both appear good, but Xcode kept yelling.

@Holzhaus I should be asking you, too, since you've been doing warnings recently. Still good?

### DIFF
--- a/core/hw/sh4/dyna/driver.cpp
+++ b/core/hw/sh4/dyna/driver.cpp
@@ -53,7 +53,7 @@ void emit_SetBaseAddr() { LastAddr_min = LastAddr; }
 void emit_WriteCodeCache()
 {
 	wchar path[512];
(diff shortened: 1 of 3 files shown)


skmp:
You 150% don't want %s there. %p would do I think

LoungeKatt:
Blame Xcode. It told me it had to be s. It was really convincing about it, too.

Holzhaus:
Looks good to me.

skmp:
Okay, i get a random crash on iOS. DEbugging

LoungeKatt:
The EXC_BAD_ACCESS code 2?

skmp:
I did some more fixing, I say we close this and resume on #766!

LoungeKatt:
Sounds like a plan

PR #762, opened 2015-08-17
Makefile: Fix Linux_x86

Holzhaus:
This is a modified version of PR #757 (which breaks ARM).

I check if it's an ARM build (NOT_ARM not defined) and then set `$(AS)` to `gcc` and append `$(CFLAGS)` to `$(ASFLAGS)`.

@skmp Is this correct? Does every ARM build use `gcc` as `$(AS)`?

If so, It *should* work fine. ~~I'm currently testing on my ODROID.~~ **EDIT:** Works fine on my ODROID-C1.

### DIFF
--- a/shell/linux/Makefile
+++ b/shell/linux/Makefile
@@ -94,7 +94,6 @@ else ifneq (,$(findstring lincpp,$(platform)))
 
 # Raspberry Pi 2
 else ifneq (,$(findstring rpi2,$(platform)))
-    AS=${CC_PREFIX}gcc
     MFLAGS += -marm -march=armv7-a -mtune=cortex-a9 -mfpu=neon -mfloat-abi=hard -funroll-loops
     ASFLAGS += -march=armv7-a -mfpu=neon -mfloat-abi=softfp
     CFLAGS += -D TARGET_BEAGLE -D TARGET_LINUX_ARMELv7 -DARM_HARDFP -fsingle-precision-constant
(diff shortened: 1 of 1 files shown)


Holzhaus:
Can I merge?

LoungeKatt:
Looks good to me. The only issue I had seen was the +=, but that's fixed now

skmp:
AS should always be GCC. Because of the preprocessing (.S file)

On Mon, Aug 17, 2015 at 8:37 PM, Jan Holthuis <notifications@github.com>
wrote:

> Merged #762 <https://github.com/reicast/reicast-emulator/pull/762>.
>
> —
> <https://github.com/reicast/reicast-emulator/pull/762#event-384546228>.
>



-- 
~skmp

Holzhaus:
Right now it isn't for non-ARM linux. However, rec-x64 doesn't have any `*.S` files and [rec-x86](https://github.com/reicast/reicast-emulator/blob/master/core/rec-x86/rec_lin86_asm.S)  does not seem need any preprocessing.

skmp:
Well, since it's *S it'd expect it to go though gcc anyway. .S files always
do (that's the diff w/ .s ones)

On Tue, Aug 18, 2015 at 4:15 PM, Jan Holthuis <notifications@github.com>
wrote:

> Right now it isn't for non-ARM linux. However, rec-x64 doesn't have any
> *.S files and rec-x86
> <https://github.com/reicast/reicast-emulator/blob/master/core/rec-x86/rec_lin86_asm.S>
> does not seem need any preprocessing.
>
> —
> <https://github.com/reicast/reicast-emulator/pull/762#issuecomment-132224485>
> .
>



-- 
~skmp

PR #763, opened 2015-08-17
README: Update 'Building for Linux' section

Holzhaus:


LoungeKatt:
Good call. I always forget we have that.

PR #764, opened 2015-08-17
linux-dist: indentation fix

Holzhaus:
This is a follow-up PR to #752. It fixes indentation (using tabs instead of spaces) like @skmp has requested :wink:


LoungeKatt:
Holy...

skmp:
@LoungeKatt this builds off #752 (forks at so only the "new" commits would diff, only after b4e46704d689b1b5d420619a59445480b00128b5)

PR #765, opened 2015-08-17
Handle .gdi files on the CLI (and other enhancements)

randomstuff:
(Updated) summary:

* `.gdi` files support as CLI arguments;
* case insensitive file extension matching (only at the end of the file name).

### DIFF
--- a/core/cfg/cl.cpp
+++ b/core/cfg/cl.cpp
@@ -127,18 +127,27 @@ bool ParseCommandLine(int argc,wchar* argv[])
 			cl-=as;
 			arg+=as;
 		}
-		else if (strstr(*arg, ".cdi") || strstr(*arg, ".chd")) {
-			printf("Using '%s' as cd image\n", *arg);
-			cfgSetVirtual("config", "image", *arg);
-		}
-		else if (strstr(*arg, ".elf")) {
-			printf("Using '%s' as reios elf file\n", *arg);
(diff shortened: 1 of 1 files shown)


randomstuff:
Should the extensions be searched at the end of the string instead of anywhere in the string?

skmp:
Is strcmpi or stricmp portable enough? The check should be case-insensitive too. Also I don't recall if you've committed stuff before, make sure to read #455 !

randomstuff:
Apparently I don't have a `strcmpi` (GNU libc). I have a POSIX `strcasecmp` but apparently it's not available on MS C runtime.

randomstuff:
* added case-insensitive matching of the file name;
* only match the end of the file name

skmp:
Lol it is? Must have done that ages ago, totally forgot about it. Can you enable travis-ci for your repo? Then it'll compile to see it all works nicely with the "offical" tools :3

randomstuff:
I tried to do that currently it's failing on the `openssl` command.

skmp:
Can you switch the logic to if/elseif/else (to grt tid of the continue). Great catch there btw!

randomstuff:
Yes, sure. Rebased and removed the continue.

skmp:
Looks sweet now! Will merge when the builds finish (or when I wake up)

PR #766, opened 2015-08-18
Fix/ios and osx

skmp:


LoungeKatt:
screen shot 2015-08-18 at 9 20 58 am

skmp:
What platform?

skmp:
(I tested it both on ios and osx and it works here)

LoungeKatt:
iOS. Are you running on an actual device?

skmp:
Yup, on my ipad. Can you try a clean compile?

skmp:
Keep it might it might not run under xcode but only when launched on it's own. xcode can't cope with some of the speed optimizations we do yet.

LoungeKatt:
Same thing. It's on an iPhone, though. Are you installing it as a system app?

skmp:
I just install it via xcode

On Tue, Aug 18, 2015 at 3:29 PM, TwistedUmbrella <notifications@github.com>
wrote:

> Same thing. It's on an iPhone, though. Are you installing it as a system
> app?
>
> —
> <https://github.com/reicast/reicast-emulator/pull/766#issuecomment-132205342>
> .
>



-- 
~skmp

LoungeKatt:
Weird. Crashes either way. I'll have to charge up my iPad and try that.

skmp:
Get on irc dammit :P

LoungeKatt:
I meant to ask you about this. The difference between the iPad and iPhone was that the phone runs 64-bit unless the app doesn't support it. This should fix it.

(Stupid app made it comment on the commit instead of the pull. MESSY!)

LoungeKatt:
@skmp After doing a lot of random checks, it seems if I define the value of CodeCache (and later ICache) in the memset function, it fixes the issue and the games boot fine.

skmp:
What do you mean?

On Wed, Aug 19, 2015 at 9:24 PM, TwistedUmbrella <notifications@github.com>
wrote:

> @skmp <https://github.com/skmp> After doing a lot of random checks, it
> seems if I define the value of CodeCache (and later ICache) in the memset
> function, it fixes the issue and the games boot fine.
>
> —
> <https://github.com/reicast/reicast-emulator/pull/766#issuecomment-132750647>
> .
>



-- 
~skmp

LoungeKatt:
Works: ```memset((u8*)mmap(ICache, ICacheSize, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_FIXED | MAP_PRIVATE | MAP_ANON, 0, 0),0xFF,ICacheSize);```

Crashes: ```memset(ICache,0xFF,ICacheSize);```

LoungeKatt:
How do I send a keycode? I have been staring at the Android and Linux versions for about an hour now and the best I can come up with is that i define kcode[], set it to a value, and it's like voodoo after that.

skmp:
Just set it, it gets read as a global variable

skmp:
You can see my "tap to button" code that does it

LoungeKatt:
Son of a... I was looking in the wrong file.

LoungeKatt:
Theoretically once the listeners are fixed, 6b56db4f6c6db76409690017ed443e98c69c1950 should work

PR #767, opened 2015-08-18
cfg: Reimplement ConfigFile using C++ maps and strings

Holzhaus:
I reimplemented the stuff from `cfg/ini.cpp` with C++ maps and string, since the original implementation was somewhat complicated to read and also kept segfaulting sometimes for no reason (I didn't find the error at least).

Bonus: It's shorter.

~~*IMPORTANT:* I currently didn't implement the `flags` stuff, so there are no virtual entries and everything is always saved. @skmp: If you want me to implement that too first, please tell me.~~ **EDIT:** Virtual entry support works now.


skmp:
Yes, flags is a must. You can override any config from the command line, and you certainly don't want that saved on the cfg file :)

core/cfg/ini.cpp
```diff
@@ -1,299 +1,252 @@
...
+		return atoi(this->value.c_str());
+	}
 }
 
-string ConfigEntry::GetValue()
+bool ConfigEntry::get_bool()
 {
-	if (flags&CEM_VIRTUAL)
-		return valueVirtual;
+	if (strcmp(this->value.c_str(), "yes") == 0 ||
```
skmp: Why not string != ? Also, maybe stricmp would make some sense here

core/cfg/ini.cpp
```diff
@@ -1,299 +1,252 @@
...
 
-			if(stricmp(c->valueVirtual.c_str(),value.c_str())==0)
-				return;
-			c->flags&=~CEM_VIRTUAL;
-		}
-	}
-	else
+bool ConfigFile::has_section(string name)
+{
+	return (this->sections.count(name) == 1);
```
skmp: extra ()s?

core/cfg/ini.cpp
```diff
@@ -1,299 +1,252 @@
...
+};
+
+bool ConfigFile::has_entry(string section_name, string entry_name)
+{
+	ConfigSection* section = this->get_section(section_name);
+	return ((section == NULL) ? false : section->has_entry(entry_name));
```
skmp: How about `section && section->has_entry(entry_name)` ?

skmp:
I've left you some  notes, looks clean and nice otherwise. Replacing code from my teenage years that has survived 5 rewrites and psp *me shivers a bit*

core/cfg/ini.cpp
Holzhaus: Do all relevant platforms support stricmp?

core/cfg/ini.cpp
Holzhaus: I like the brackets. Do you want me to remove them?

Holzhaus:
@skmp I added the virtual entry support.

randomstuff:
I have a regression with this PR in commit "b2360c8 cfg: Reimplement ConfigFile using C++ maps and strings" (as well as in the evdev-mappings-from-file branch in "e022d90 cfg: Reimplement ConfigFile using C++ maps and strings"): it's not loading the image given in CLI argument anymore.

core/cfg/ini.cpp
```diff
@@ -1,299 +1,280 @@
...
+
+			char* name = trim_ws(tl);
+			char* value = trim_ws(separator + 1);
+			if (name == NULL || value == NULL)
 			{
-				ConfigSection*cs=this->GetEntry(cur_sect);
-				
-				//if (!cs->FindEntry(k))
-				cs->SetEntry(k,v,CEM_SAVE|CEM_LOAD);
+				printf("Malformed entry on config - ignoring @ %d(%s)\n",cline, tl);
```
randomstuff: An mepty size if recognised as an error. This is consistent with the previous behaviour but I wonder if this what we want especially since there is this code:

~~~c
settings.reios.ElfFile = cfgLoadStr("reios", "ElfFile","");
~~~

which initialised this as an empty string by default.

randomstuff:
My issue (cannot choose the image from the CLI) is related to the lack of virtual config I guess:

~~~c
else if (strstr(*arg, ".cdi") || strstr(*arg, ".chd")) {
  printf("Using '%s' as cd image\n", *arg);
  cfgSetVirtual("config", "image", *arg);
}
~~~

If I add a entry:

~~~
[config]
image=/path/too/foobar.cdi
~~~

I can load the image.

Holzhaus:
@randomstuff Have a look at 52e98eb. Passing the image via CLI works for me.

skmp:
This really exemplifies why we need tests. HRFM.

Holzhaus:
Well, he tested the commit before I added the virtual flags stuff

randomstuff:
OK, I missed the message when you said: "I added the virtual entry support." so I tested it witout it. :/

Holzhaus:
@skmp Can we merge or does this need more testing?

skmp:
I want to do a second read-through the code to make sure I have a good grasp and nothing is missing, and ill +1 then. gimmeh a day :)

Holzhaus:
OK :D

skmp:
Okay, testing on windows I get "Inconsistent Stream Count. Flush between consecutive read and write"

skmp:
Which makes sense, because `FILE* cfgfile = fopen(cfgPath.c_str(),"r");` doesn't go too well with

```
	if (cfgfile)
	{
		cfgdb.save(cfgfile);
		fclose(cfgfile);
	}
```

Holzhaus:
@skmp: Strange that you didn't encounter this before though. It was one of the parts [i didn't change](https://github.com/reicast/reicast-emulator/blob/6348057402c59575f02d621001ac413750b8d0f9/core/cfg/cfg.cpp#L292-L367).

Anyhow, I removed that bit, since it's useless to save the file immediately after parsing it. The only thing that happened between parsing and saving was setting some virtual values, which didn't affect the output of the saved file anyway, so this resulted in unnecessary writes.

New behaviour: Only save if some value have been changed ;-)

Please retest.

skmp:
Yeah, was digging a bit into it.. wth?
Re testing!

skmp:
I guess the save was there to create the cfg if it was just created. There used to be a comment line on top like ";nulldc configuration" or whatever

skmp:
Rebased to master to test

skmp:
```
	cfgPath=GetPath("/emu.cfg");
	FILE* cfgfile = fopen(cfgPath.c_str(),"r");
```

Now it won't create the file at all :p

skmp:
Looks to work nicely now on windows! Can you check on linux and merge if it's good?

Holzhaus:
I pushed some more fixes, can you retest (and merge if that works for you)?
One fixes a small inconsistency (`cfgSaveStr` also writes the config file, but `cfgSaveInt` didn't do that), the other one (f56299209dd5854a38320eee445a243dda6a84df) only creates the the new config file if it didn't exist before (and does not create it if there was a permissions problem. It also disables saving via `cfgSaveStr` in that case).

skmp:
All looks good. I'm still not convinced that the erro logic is worth the effort, but since you wrote it already ... :p

PR #772, opened 2015-08-19
Remove unnecessary includes

Holzhaus:
Are these needed? I was able to compile without them.

### DIFF
--- a/core/hw/pvr/drkPvr.cpp
+++ b/core/hw/pvr/drkPvr.cpp
@@ -16,7 +16,6 @@
 #include "pvr_regs.h"
 #include "pvr_mem.h"
 #include "Renderer_if.h"
-#include <algorithm>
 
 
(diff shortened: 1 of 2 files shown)


Holzhaus:
@skmp Are these still needed or can I merge?

skmp:
Possibly not. They were included for sort() most likely, but then refactor+++. Is there any #ifdef madness going on on the related files?

skmp:
Ahh, now this is a good example where my "don't change unless an issue" ethic comes from. In C++11 `std::swap` was moved from `algorythm` to `utility`. We still support C++98 compilers (and I'd love to avoid depending on C++11 on the core for the foreseeable future, at least when it's easy. It makes portability harder). `swap` is used in `descrambl.cpp`

Holzhaus:
I'll close this then.

PR #774, opened 2015-08-19
Linux: Enable x11 fullscreen toggling and fix window resizes

### LINKED ISSUE #761: [Linux] X11 window doesn't resize properly
Currently, x11 window resizing isn't implemented yet.

I'm able to catch window resize events and query the new size of the window, but since I don't have any OpenGL experience, I don't know how to make the video output fill the new window size.

Holzhaus:
This one enables toggling fullscreen with the F11 key and also fixes window resizing. Resolves #761.

For avoiding merge conflicts, we need to merge #752 & #764 first.

### DIFF
--- a/core/linux-dist/x11.cpp
+++ b/core/linux-dist/x11.cpp
@@ -15,11 +15,13 @@
 #include "linux-dist/main.h"
 
 #if defined(TARGET_PANDORA)
-	#define WINDOW_WIDTH  800
+	#define DEFAULT_FULLSCREEN    1
+	#define DEFAULT_WINDOW_WIDTH  800
 #else
-	#define WINDOW_WIDTH  640
+	#define DEFAULT_FULLSCREEN    0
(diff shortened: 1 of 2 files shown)


meveric:
once again.. having a button as the only method to switch to full screen mode can be troublesome..
Imagine a scenario with a device running as a console with only Joystick control but no keyboard connected.
It would be better to have a config option as well to switch to full screen mode, rather than only rely on the keyboard alone.

Holzhaus:
Sure I'll add that when I have the time, but being able to fullscreen the app at all was a priority ;-)

Holzhaus:
@meveric commit 4dee565 should make you happy :D

skmp:
Looks good to me. Finally proper x11 fullscreen support :+1:

I know formatting is important, but let's try to keep the formatting changes on separate PRs. It helps when I read the diffs (and I read a ton of diffs these days -.-)

PR #782, opened 2015-08-20
[Wercker] Add preliminary wercker.yml for Linux_x64 building

Holzhaus:
This will build reicast for Linux x64 on Ubuntu. ARM/x86 support is still missing though.

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -1,0 +1,12 @@
+box: ubuntu
+
+build:
+    steps:
+        - install-packages:
+            packages: build-essential libasound2-dev libgl1-mesa-dev libx11-dev
+        - script:
+            name: gcc-version
+            code: gcc --version
+        - script:
+            name: reicast x64 build
+            code: make -C shell/linux platform=x64


Holzhaus:
I'll just merge this for now, feel free to improve the docker.yml (probably by using a dedicated docker container, so that we don't need to install all the dependencies every time).

skmp:
Woohoo, and linux CI :D

ship.io also builds for ios now too. I doesn't report to github (yet?) and it's a single user (we'll prolly update to a multi-user license whenever nillware (I) can afford it :p). Now we need a "mailing-list-as-a-service" thingy to handle the mail load as a team too. 

/me does the happy dance

Holzhaus:
How about google groups?

skmp:
I dislike them. I want to keep github as the cetral place of coordination and discussion, and irc for more detailed matters -- keeps the history of the project on a single project. (Now, only if git had build in support for issues, that would be awesome.. no?)

I just want to distribute user feedback and build failures, etc

PR #783, opened 2015-08-21
linux-dist: Remove unneeded code

Holzhaus:
Do we need the curses code? I can't see it serve any particular purpose right now. Might be a leftover from ages ago.

### DIFF
--- a/core/linux-dist/main.cpp
+++ b/core/linux-dist/main.cpp
@@ -38,7 +38,7 @@
 #ifdef TARGET_PANDORA
 	#include <signal.h>
 	#include <execinfo.h>
-	#include <sys/soundcard.h>  
+	#include <sys/soundcard.h>
 #endif
 
 
@@ -97,10 +97,10 @@ void SetupInput()
 	#if defined(USE_EVDEV)
 		int evdev_device_id[4] = { -1, -1, -1, -1 };
 		size_t size_needed;
-		
+
 		int evdev_device_length, port, i;
 		char* evdev_device;
-		
+
 		for (port = 0; port < 4; port++)
 		{
 			size_needed = snprintf(NULL, 0, EVDEV_DEVICE_CONFIG_KEY, port+1) + 1;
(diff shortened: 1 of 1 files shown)


skmp:
Yes, we do. I use it when testing over ssh to some devboard (very common for my setup)

Holzhaus:
Hmm, it works fine for me without it (ran it via ssh). Anyway, if you need it, can't you just run `stty echo -icanon`?

skmp:
IIRC the point of the code is for non-blocking stdin. Ill double check it when on my lappy

Holzhaus:
Non-blocking stdin should be possible via:

```
stty echo -icanon time 0 min 0
```

skmp:
I haven't used stty before, and it looks like it would work, but still I prefer reicast setting it up. It's reicast's responsibility to setup the input to a sate that it works for itself (and to cleanup afterwards, which atm doesn't happen because of unclean exits)

Holzhaus:
Where does reicast read from stdin though?

skmp:
Hum, it used to be on [man.cpp](https://github.com/reicast/reicast-emulator/blob/a90a2cf91c1f2329ae5ea182d7283a3825ce1133/core/linux-dist/main.cpp#L457). I don't know where this code resides atm though

skmp:
Hum, looks like you got rid of the code on the evdev changes. Hrfm.

Holzhaus:
You may slap me in the face now. I removed it in 14ea9e4196f763638ce3f206c74adb69af610c22. I'll readd `stdin` input in a separate PR, ok?

skmp:
Nah don't worry too much about it. Apart form me I'm not sure anyone else uses that part. Since the input stuff is getting so much re-work let's get rid of this and I can re-add next time I need in a cleaner way.

skmp:
Wthat's `sys/soundcard.h` used for though?

skmp:
Oh that was an unrelated whitespace change. k, merging...

PR #789, opened 2015-08-21
msvc: Update sln files to use "x86" instead of "win32" for appveyor

skmp:
Getting everything working for appveyor, building all 4 configurations

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,9 @@
+version: git-{branch}-{build}
+configuration: Debug
+platform:
+- x86
+- x64
+build:
+  project: shell/reicast.sln
+  parallel: true
+  verbosity: minimal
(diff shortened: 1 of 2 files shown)


skmp:
Rebased to msater for Wercker fixes ~

skmp:
Okay, building release for windows takes a while (~ 30 mins, because of enabled global optimizations). I'm thinking of doing debug only builds for now, and enable release building for when we upgrade to a plan w/ more beefy cpus!

skmp:
Okay, debug builds take 1 min each :d Still queing time can be quite high, but wooohoo!!!

Holzhaus:
Nice work :D

harissut:
ok im waiting this windows x86/x64.thanx

PR #790, opened 2015-08-21
Add a generic 32-bit ARM rule.

sergiobenrocha2:
It can be useful in places like launchpad farm.

### DIFF
--- a/shell/linux/Makefile
+++ b/shell/linux/Makefile
@@ -56,7 +56,7 @@ ifeq (,$(platform))
         else ifneq (,$(findstring ODROIDU2,$(HARDWARE)))
             platform = odroidu2
         else
-            platform = arm
+            platform = armv7h
         endif
     else ifneq (,$(findstring mips,$(ARCH)))
         platform = gcwz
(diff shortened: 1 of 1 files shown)


sergiobenrocha2:
@skmp are there anything wrong with this commit?

sergiobenrocha2:
Many boards nowadays comes with neon support in the SoC

skmp:
I think what would be best is to auto-detect the current platform,

`uname -m` works for thearchitecture (eg, arm7l) and `ls /lib/* | grep armhf` to detect armhf or not. Reicast requires at least neon-vfp16. You can do `cat /proc/cpuinfo | grep Features` which gets us a list like `half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm`. Reicast requires `vfpv3` or `vfpv4` and `neon`. `vfpd32` is optional (the JIT doesn't use the extra registers, but clang/gcc might if you enable it)

skmp:
Launchpad does cross compiles though? Detection logic only helps for when target and build machine are the same

sergiobenrocha2:
Launchpad uses Qemu to build ARM packages.

skmp:
Qemu *URGH*

skmp:
(I still haven't made up my mind if we should default to arm or not, which is why I haven't merged this in yet)

Holzhaus:
@skmp Actually, we woudn't default to ARM in all cases - just if it's actual ARM hardware (and thus `$ARCH` contains the string `arm`: https://github.com/sergiobenrocha2/reicast-emulator/blob/master/shell/linux/Makefile#L36-L60

skmp:
Hah, I failed to notice that part. Yeh, that makes perfect sense.

PR #791, opened 2015-08-21
Update libpng to 1.6.18.

sergiobenrocha2:
It's working on linux x86-64


skmp:
It workd against the global zlib, not the local one (thats why CI build fails)

skmp:
:+1: from me. @LoungeKatt can you check it works on android? My build ndk env is broken again -.-

skmp:
@LoungeKatt did you had any opportunity to check this? (/me is in let's merge everything mode)

PR #794, opened 2015-08-22
Update zlib to 1.2.8.

sergiobenrocha2:
It's working on linux x86-64


skmp:
Looks like the version we were using was rather old. I'll do a couple of tests tomorrow to make sure everything works as expected, and then merge. Meanwhile, pleas read over #455 ~

skmp:
Okay, seems to be working alright!

PR #797, opened 2015-08-22
nulldc: Remove some hardcoded values

Holzhaus:
According to the search function, `OMAP4` isn't set anywhere, so this shouldn't be a problem.

The other stuff also doesn't belong there IMHO. Guess this was testing code that hasn't been removed yet.


skmp:
yeh, leftovers from another age

PR #769, opened 2015-08-18
Read evdev mappings from file

Holzhaus:
This is needs PRs #752, #764 and #767, so it's actually starting with commit ~~a69f85b~~ (**EDIT:** I rebased, it starts with ee6fca7).

This is somehow big nonetheless, although there are not that many code changed. The biggest thing is the new `core/linux-dist/evdev_keycodes.cpp` (660 lines), so that uses can do `btn_a = BTN_A` instead of `btn_a = 0x130` or `btn_a = 304`. It sucks, I know, but I think it's worth it.

PR #798, opened 2015-08-22
Introduce separate config/data dirs (user & system wide)

Holzhaus:
This one has grown a bit bigger than I expected. Anyhow, I here's a description what this PR does:

It replaces the `GetPath` function with two dedicated functions:

```C++
get_config_path(filename);
get_data_path(filename);
```

The first one gets a path for a config file (we currently only have `emu.cfg`). The second one gets a path for a data file (e.g. `buttons.png`, the VMU saves or the BIOS/Flashrom).

I did this separation, because on some platforms, the stuff is actually saved in different places - on Linux, this is `$HOME/.config/reicast` for config stuff and `$HOME/.local/share/reicast` for data files (see [XDG Basedir Spec](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)).

I also added support for distinguishing writable and read-only files. Both functions take a second parameter (`bool user_writable`). This one is `true` by default.
- If `user_writable` is `true`, it'll always return a user-writable filepath (i.e. 
`~/.config/reicast/somefile.txt` and `~/.local/share/reicast/somefile.txt`, respectively).
- If that variable is `false` and `~/.config/reicast/somefile.txt` / `~/.local/share/reicast/somefile.txt` does not exist, it also checks system-wide dirs (e.g. `/usr/share/reicast` for data files on Linux).

That way, a user can always override the default/system-wide files, but he doesn't have to duplicate them inside the home dir if he doesn't want to.

On all other platforms (e.g. Windows), i just replaced `SetHomeDir(foo)` with:

```C++
set_user_config_dir(foo);
set_user_data_dir(foo);
```

Also, system-wide directories won't be searched for now - but that could easily be added by using:

```C++
add_system_config_dir(foo);
add_system_data_dir(foo);
```

**Example:**
We merged #769 and a user has just installed reicast (either by self-compilation or by installing via package manager).

Input won't work out of the box, because the user needs to copy the `shell/linux/mappings` folder into `~/.reicast` first (even the pre-defined mappings won't work until this happens).

With this PR, it's possible to put the pre-defined mappings into `/usr/share/reicast/mappings`.

When calling `get_data_path("somemapping.cfg")`, reicast will look in these folders:
- `~/.local/share/reicast/somemapping.cfg`
- `/usr/share/local/reicast/somemapping.cfg`
- `/usr/share/reicast/somemapping.cfg` (in that order)

It'll return the first filepath that exists or - if none of the files exist - the user-dir (i.e. `~/.local/share/reicast/somemapping.cfg`).

### DIFF
--- a/shell/apple/emulator-osx/emulator-osx/osx-main.mm
+++ b/shell/apple/emulator-osx/emulator-osx/osx-main.mm
@@ -95,10 +95,14 @@ void* emuthread(void*) {
     {
         home += "/.reicast";
         mkdir(home.c_str(), 0755); // create the directory if missing
-        SetHomeDir(home);
+        set_user_config_dir(home);
+        set_user_data_dir(home);
     }
     else
-        SetHomeDir(".");
+    {
+        set_user_config_dir(".");
+        set_user_data_dir(".");
+    }
     char* argv[] = { "reicast" };
     
     dc_init(1,argv);
--- a/core/cfg/cfg.cpp
+++ b/core/cfg/cfg.cpp
@@ -78,9 +78,11 @@ void  cfgSaveStr(const wchar * Section, const wchar * Key, const wchar * String)
 
 bool cfgOpen()
 {
-	cfgPath=get_config_path("/emu.cfg");
-	FILE* cfgfile = fopen(cfgPath.c_str(),"r");
+	const char* filename = "/emu.cfg";
+	string config_path_read = get_readonly_config_path(filename);
+	cfgPath = get_writable_config_path(filename);
 
+	FILE* cfgfile = fopen(config_path_read.c_str(),"r");
 	if(cfgfile != NULL) {
 		cfgdb.parse(cfgfile);
 		fclose(cfgfile);
@@ -89,8 +91,8 @@ bool cfgOpen()
 	{
 		// Config file can't be opened
 		int error_code = errno;
-		printf("Warning: Unable to open the config file '%s' for reading (%s)\n", cfgPath.c_str(), strerror(error_code));
-		if (error_code == ENOENT)
+		printf("Warning: Unable to open the config file '%s' for reading (%s)\n", config_path_read.c_str(), strerror(error_code));
+		if (error_code == ENOENT || cfgPath != config_path_read)
 		{
 			// Config file didn't exist
 			printf("Creating new empty config file at '%s'\n", cfgPath.c_str());
--- a/core/cfg/cfg.cpp
+++ b/core/cfg/cfg.cpp
@@ -78,7 +78,7 @@ void  cfgSaveStr(const wchar * Section, const wchar * Key, const wchar * String)
 
 bool cfgOpen()
 {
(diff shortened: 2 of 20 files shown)


randomstuff:
I guess VMU saves and stuff like data/buttons.png should go in $XDG_DATA_HOME?

Holzhaus:
Yes, but I'm still working on it (Going to rewrite the `GetPath` logic).

Holzhaus:
@randomstuff @skmp Check out the updated PR description.

skmp:
Just arived at Berlin, will take a look on this and all other PRs in the
afternoon :)
Am 23.08.2015 01:04 schrieb "Jan Holthuis" <notifications@github.com>:

skmp:
Hmm, the file overlays make it interesting. I think it would be best to have different functions for read and write dirs -- since we're making the source easier to read overall anyway. Since we don't have an official style guide -yet- i think it makes more sense for the names to be like GetConfig[Read/Write]Path to preserve the style of the GetPath function they replace. 

~

Holzhaus:
That is already there: `get_config_path` take a second parameter which is true by default and indicates if you intent to write to that path.

skmp:
Yeh, but I'm not sure this is clear enough. I'd rather have the function
names a bit closer to what they do (don't take the existing source as an
example, it sucks). Eg, FindReadPath might be even better.
Am 24.08.2015 14:57 schrieb "Jan Holthuis" <notifications@github.com>:

Holzhaus:
How about `get_user_config_path` and `get_system_config_path`?

*Note to myself:* I probably shouldn't [edit](https://github.com/reicast/reicast-emulator/pull/798#issuecomment-134188189) my previous comments to add information.

Holzhaus:
@skmp I rebased and renamed them to `get_writable_data_path()` and `get_readonly_data_path()`. Ok?

skmp:
I'm again running short on time, will review and merge things on sunday.
And also work on the crowdfund video.

On Sat, Aug 29, 2015 at 1:34 AM, Jan Holthuis <notifications@github.com>
wrote:

> @skmp <https://github.com/skmp> I rebased and renamed them to
> get_writable_data_path() and get_readonly_data_path(). Ok?
>
> —
> <https://github.com/reicast/reicast-emulator/pull/798#issuecomment-135913361>
> .
>



-- 
~skmp

shell/apple/emulator-osx/emulator-osx/osx-main.mm
```diff
@@ -95,10 +95,12 @@ void gl_swap() {
     {
         home += "/.reicast";
         mkdir(home.c_str(), 0755); // create the directory if missing
-        SetHomeDir(home);
+        set_user_config_dir(home);
+        set_user_data_dir(home);
     }
     else
-        SetHomeDir(".");
+        set_user_config_dir(".");
```
skmp: Missing {}s ?

skmp:
Apart from the osx-main thingy, i'm :+1: on this

PR #807, opened 2015-08-25
Nixprof: some enchancements

randomstuff:
* Make it available on X11
* Make it work with ELF64
* Use .symtab if available instead of .dynsym
* Only process STT_FUNC symbols
* Use section types and links insteaf of section names to find symbol and string table sections

Some modifications to deps/libelf. Is this code maintained at https://github.com/seL4/libelf? Needs to be upstreamed.

Might be a good idea to use perf or directly perf_event_open (faster, less intrusive, lots of features). For the former, a /tmp/perf-$PID.map can be used to bind SH4 blocks to host addresses.

### DIFF
--- a/core/linux-dist/x11.cpp
+++ b/core/linux-dist/x11.cpp
@@ -14,6 +14,10 @@
 #include "linux-dist/x11.h"
 #include "linux-dist/main.h"
 
+#if FEAT_HAS_NIXPROF
+#include "profiler/profiler.h"
+#endif
+
 #if defined(TARGET_PANDORA)
 	#define DEFAULT_FULLSCREEN    1
 	#define DEFAULT_WINDOW_WIDTH  800
@@ -72,6 +76,16 @@ void input_x11_handle()
 					{
 						die("death by escape key");
 					}
+#if FEAT_HAS_NIXPROF
+					else if (e.type == KeyRelease && e.xkey.keycode == 76) // F10 button
+					{
+						if (sample_Switch(3000)) {
+							printf("Starting profiling\n");
+						} else {
+							printf("Stopping profiling\n");
+						}
+					}
+#endif
 					else if (e.type == KeyRelease && e.xkey.keycode == 95) // F11 button
 					{
 						x11_fullscreen = !x11_fullscreen;
--- a/core/deps/libelf/elf.cpp
+++ b/core/deps/libelf/elf.cpp
@@ -229,7 +229,7 @@ elf_getSectionLink(void *elfFile, int i)
 {
 	return ISELF32 (elfFile)
 		? elf32_getSectionLink((Elf32_Header*)elfFile, i)
-		: 0;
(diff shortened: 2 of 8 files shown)


Holzhaus:
Looks reasonable to me, although I don't really know what that code does, so @skmp's opinion will be more valueable here.

skmp:
There's no race condition (only single writer), but the design isn't the best one. I'm not sure nixprof still works alright though. There's a C#/windows front end that can visualize the data in a pretty nice way. I wanted to rewrite it to html5 + websockets. Another thing on the wishlist....

randomstuff:
After the fixes, it seems to be working OK.

Holzhaus:
@randomstuff can you rebase? I guess commit 4890e59 has a merge conflict with def93ca.

randomstuff:
Rebased the code.

randomstuff:
I'm probably going to try to implement `/tmp/perf-$pid.map` support which would allow to do basically the same thing with `perf` or the [poor man's profiler](http://poormansprofiler.org/) + some tooling.

skmp:
nixprof was (is?) unique in that it stamped most of the samples and it
works on android (without messing with the kernels). I don't know the state
of things now, though.

On Fri, Aug 28, 2015 at 8:59 PM, Gabriel Corona <notifications@github.com>
wrote:

> I'm probably going to try to implement /tmp/perf-$pid.map support which
> would allow to do basically the same thing with perf or the poor man's
> profiler <http://poormansprofiler.org/> + some tooling.
>
> —
> <https://github.com/reicast/reicast-emulator/pull/807#issuecomment-135860706>
> .
>



-- 
~skmp

randomstuff:
Yes, I don't known how easy it is to make external profiling tools on Android. Apparently, the kernel already has support for perf event (?). Perf timestamps the samples as well but it cannot easily sample [off-cpu](http://www.brendangregg.com/blog/2015-02-26/linux-perf-off-cpu-flame-graph.html) along with on-cpu time.

LoungeKatt:
They're not reliable.

core/linux-dist/x11.cpp
```diff
@@ -72,6 +76,16 @@ void input_x11_handle()
 					{
 						die("death by escape key");
 					}
+#if FEAT_HAS_NIXPROF
+					else if (e.type == KeyRelease && e.xkey.keycode == 76) // F10 button
+					{
+						if (sample_Switch(97)) {
```
skmp: 97 hz is wayy too low to get accurate profiling results. I originally used 1K to 10K

skmp:
Well, seeing this breaks nothing, and the changes are minor and in-style, i'm :+1: Check the hz comment. I think I use 3K on android.

core/linux-dist/x11.cpp
randomstuff: OK, I'll change this.

randomstuff:
Changed the sampling frequency and rebased against master.

Issue #800, opened 2015-08-23
In appveyor buildbot doesn't present build result (*.zip with *.exe etc).


ofry:
For sample, https://ci.appveyor.com/project/skmp/reicast-emulator/branch/master/job/l0ul3u5ydwy91af0/artifacts

The build job does not contain any artifacts.

ofry:
Thanks.

Holzhaus:
I'll write some api parser to add the URLs to the reicast ci builds page.

Holzhaus:
Damn, that doesn't work. AppVeyor doesn't have a JSONP API.

skmp:
Just FYI, http://builds.reicast.com now has the windows zips

PR #810, opened 2015-08-25
AppVeyor: Add build artifacts

Holzhaus:
We might want to budle the exe with some "Ask Toolbar" installer, since
Windows users might be confused if it's not.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,8 +11,8 @@ build:
 after_build:
   - cmd: mkdir artifacts
   - cmd: set PLATFORM2=%PLATFORM:x86=Win32%
-  - cmd: move WorkDir\reicast_%PLATFORM2%_Slow.exe artifacts/reicast-%PLATFORM%-%APPVEYOR_REPO_COMMIT%.exe
+  - cmd: move WorkDir\reicast_%PLATFORM2%_Slow.exe artifacts/reicast-%PLATFORM%-%APPVEYOR_REPO_COMMIT%-slow.exe
 
 artifacts:
   - path: artifacts
-    name: reicast-$(platform)
+    name: reicast-$(platform)-slow
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,3 +7,12 @@ build:
   project: shell/reicast.sln
   parallel: true
   verbosity: minimal
+
+after_build:
+  - cmd: mkdir artifacts
+  - cmd: set PLATFORM2=%PLATFORM:x86=Win32%
+  - cmd: move WorkDir\reicast_%PLATFORM2%_Slow.exe artifacts/reicast-%PLATFORM%-%APPVEYOR_REPO_COMMIT%.exe
+
+artifacts:
+  - path: artifacts
+    name: reicast-$(platform)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,8 +11,8 @@ build:
 after_build:
   - cmd: mkdir artifacts
   - cmd: set PLATFORM2=%PLATFORM:x86=Win32%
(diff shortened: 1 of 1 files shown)


Holzhaus:
By the way, this resolves #800.

skmp:
We are building the debug builds, so not sure how useful they are. I'd keep the slow modifier (it indicates debug build)

Holzhaus:
Fixed.

Holzhaus:
Maybe we should also build the non-debug ones then (or we don't, since these are builds for testing anyway).

We should probably push them to S3 and list them on http://builds.reicast.com/ for easy downloading. Right now I can't figure out an easy way to directly link to the latest build artifacts of the master branch. The casual windows user isn't used to compile stuff himself in my experience.

skmp:
But for windows, reicast isn't a wise choice atm anyway. We don't want
having to support a very non-user friendly build on such a huge
nontechnical audience. Crap from pissed android users is enough for me :p.

Yes, i want to update the builds page, but I gotta figure out how to store
that stuff. The current 'no-backend' amazon solution might not be up to the
task.

Release builds use global LTGC. This gets us wickedly fast C++ code, but
takes 5 mins or so to build on my laptop (partly because of the softrend
generating 1k functions). Appveyor free vms are quite weak too, leading to
30+ min build time for all 4 configurations. I don't think it's worth the
wait to build rls while we're on the free plan. The open source premium
plan has 3-10x faster builds, I'm told.
Am 25.08.2015 08:48 schrieb "Jan Holthuis" <notifications@github.com>:

skmp:
this is perfect :D. Merging now :)

Holzhaus:
@skmp I was sure I'd be able to add the links to the CI builds page, but I can't - since AppVeyor's API only supports JSON, not JSONP.

Should I add a deploy step for amazon? I [added a warning](https://github.com/reicast/reicast-builds/commit/8d64b6060dc8c73a502d8d67b6c9cc4ed977b912) to the CI page, so users should be upset if something doesn't work (they might still be though^^).

skmp:
appveyor should have some CORS setup for json - jsonp is just an old hack nowdays. I'll add the reicast S3 deploy keys

skmp:
Also, from android we deploy to "{branch}-{comitHash}/{reicast-name}-{platform}-{buildConfig}-{comitHash}.apk". I'll see if we can get the same structure for windows...

skmp:
@Holzhaus I did some more improvements on #811 (which I'll re-write into a single nice commit). 

Amazon Bucket: reicast-builds-windows
Should have the same format as the .apk one (reicast-builds)

`$.get('https://reicast-builds-windows.s3.amazonaws.com/').done(function(response) { console.log(arguments); });` works :dancer:

It's gonna be a bit of a meh to merge the artifacts, cus different hash lenghts, but oh well :3

skmp:
Users will always be upset. That's why they're users. It's our duty to make
their lives a bit less better without becoming miserable in the process :3

On Wed, Aug 26, 2015 at 1:53 AM, Jan Holthuis <notifications@github.com>
wrote:

> @skmp <https://github.com/skmp> I was sure I'd be able to add the links
> to the CI builds page, but I can't - since AppVeyor's API only supports
> JSON, not JSONP.
>
> Should I add a deploy step for amazon? I added a warning
> <https://github.com/reicast/reicast-builds/commit/8d64b6060dc8c73a502d8d67b6c9cc4ed977b912>
> to the CI page, so users should be upset if something doesn't work (they
> might still be though^^).
>
> —
> <https://github.com/reicast/reicast-emulator/pull/810#issuecomment-134772861>
> .
>



-- 
~skmp

PR #817, opened 2015-08-27
Linux/Makefile: Add preliminary NAOMI support

### LINKED ISSUE #795: Naomi support for linux makefile
@Holzhaus maybe you can take a for this? it needs to -D TARGET_NAOMI and change binary to reicast_naomi.elf

Also, I'm not sure how much sense the .elf extension makes for the Linux crowd. It makes a ton of sense for me, but I've used .elfs a ton for dc/ps2/wii development and I thin the typical thing in linux is to not keep the extension. Could we also do a make install ?

Holzhaus:
@skmp This should resolve #795 for the moment.


skmp:
As the obj files collide, can you make it obj-{naomi/dreamcast}-{platform} ? Looks good otherwise :+1:

PR #818, opened 2015-08-27
sh4/mmu: Import the old mmu implementation from nullDC.

skmp:
Reicast doesn't support exceptions yet, so this isn't of much use now,
and is intended mostly as documentation. nullDC used some call stack
hooking magic to handle exceptions, which was never generic and clean
enough to be worth the effort to port to Reicast.

### DIFF
--- a/shell/reicast.vcxproj
+++ b/shell/reicast.vcxproj
@@ -288,8 +288,8 @@
     <ClInclude Include="..\core\hw\sh4\interpr\sh4_opcodes.h" />
     <ClInclude Include="..\core\hw\sh4\modules\ccn.h" />
     <ClInclude Include="..\core\hw\sh4\modules\dmac.h" />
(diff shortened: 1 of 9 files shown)


shell/reicast.vcxproj
```diff
@@ -288,8 +288,8 @@
     <ClInclude Include="..\core\hw\sh4\interpr\sh4_opcodes.h" />
     <ClInclude Include="..\core\hw\sh4\modules\ccn.h" />
     <ClInclude Include="..\core\hw\sh4\modules\dmac.h" />
-    <ClInclude Include="..\core\hw\sh4\modules\intc.h" />
```
Holzhaus: Was this removed by accident? Or did you remove it on purpose?

skmp: This one was removed in 2007. The file reference somehow still survived (and header files don't get compiled so no compile error)

PR #820, opened 2015-08-28
Move settings loading out of NullDC

Holzhaus:
I'd really like to move the code that loads settings out of `dc_init()`, because i'd like to modify settings after loading but before initializing the emulator.

One reason for that is that I'd like to add multiplayer support soon. The platform-specific code can detect which controllers are plugged in (or configured) and then enable or disable the respective maple ports. This has to happen after loading the config file (obviously), but _before_ calling `dc_init()`. Something like:

``` C++
settings_t dc_settings;
dc_get_settings(&dc_settings, argc, argv);
//now I can do something like this:
dc_settings.maple.port1 = evdev_controller1_connected;
dc_settings.maple.port2 = evdev_controller2_connected;
dc_settings.maple.port3 = evdev_controller3_connected;
dc_settings.maple.port3 = evdev_controller4_connected;
dc_init(dc_settings);
```

Sure, you have to create the settings object inside the platform code, but this is merely two additional lines and IMHO much cleaner than parsing the config file and the command line inside `dc_init()`. Also, this makes `dc_init()` more predictable if we ever wanted to do unit testing.

For backwards compatiblity, there's still a `dc_init()` that takes `argc`/`argv` parameters and can be used just like before, so if you don't need it, don't use it ;-)

Alternative to my approach would be:
- Hardcode number of ports for all platforms and setups (not cool).
- Add `#ifdef`'d code for all platforms to the `maple` stuff, so that port settings are deduced from the platform-specific config entries concering input, but that solution sucks IMHO.
- Add new config entries like:
  
  ``` ini
  [config]
  maple.port1 = 1
  maple.port2 = 1
  maple.port3 = 0
  maple.port4 = 0
  ```
  
  Which is quite user-unfriendly, since we have to define even more config keys although we could auto-detect which controllers are actually connected/configured. On Linux, if I wanted to add a new controller for player 2, I'd have to specify the `evdev_device_id_2` and set `maple.port2` to `1` instead just doing the former.
- Call some kind of extern `setup_input`  function from the maple code that does this automatically and is defined in the plaform code. I really don't like this kind of tight coupling and this really makes unit-testing somewhat impossible. I understand why we have to do this in some places for performance reasons but we should keep the number of code parts doing this as small as possible.
- _Placeholder for some method i might have forgotten_

@skmp What do you think?


skmp:
I'm still wayy too tired to make a constructive comment about this. Been a busy week :|

skmp:
I guess more like a hyper-busy 3 months. I see your point now. Hrfm.

skmp:
I see you're trying to get rid of the lovely globals -- but the cmd line parsing depends internally on some of them (virtual config loading)

skmp:
(closing this as it is stale)

PR #823, opened 2015-08-28
Some groundwork for multiplayer support

Holzhaus:
~~This PR is still a work in progress, so expect rebasing/force-pushing to this branch.~~ If you have any concers about this, you can comment here :D

What this does:
- Removes 113 lines of duplicated code
- Makes it possible to enable specific ports (like port 1 and 3, but not 2 and 4). Current behaviour remains unchanged (only port 1 is enabled at the moment)
- Groups `kcode`, `joyx`, `joyy`, `lt` and `rt` together (since they belong together anyway) and gives them a more meaningful name

This might be useful anyway, but if we also merged #820, we could enable/disable ports before calling `dc_init()` based on what devices are configured and connected to the host OS.

### DIFF
--- a/core/hw/maple/maple_cfg.cpp
+++ b/core/hw/maple/maple_cfg.cpp
@@ -3,6 +3,7 @@
 #include "maple_helper.h"
 #include "maple_devs.h"
 #include "maple_cfg.h"
+#include "maple_controller.h"	
 
 #define HAS_VMU
 /*
@@ -21,11 +22,6 @@ Plugins:
 		ImageUpdate(data);
 */
 void UpdateInputState(u32 port);
-
-extern u16 kcode[4];
-extern u32 vks[4];
-extern s8 joyx[4],joyy[4];
-extern u8 rt[4],lt[4];
 
 u8 GetBtFromSgn(s8 val)
 {
--- a/core/hw/maple/maple_cfg.cpp
+++ b/core/hw/maple/maple_cfg.cpp
@@ -41,11 +41,11 @@ struct MapleConfigMap : IMapleConfigMap
 	{
 		UpdateInputState(dev->bus_id);
 
-		pjs->kcode=kcode[dev->bus_id] | 0xF901;
-		pjs->joy[PJAI_X1]=GetBtFromSgn(joyx[dev->bus_id]);
-		pjs->joy[PJAI_Y1]=GetBtFromSgn(joyy[dev->bus_id]);
-		pjs->trigger[PJTI_R]=rt[dev->bus_id];
-		pjs->trigger[PJTI_L]=lt[dev->bus_id];
+		pjs->kcode           = maple_controller[dev->bus_id].buttons | 0xF901;
+		pjs->joy[PJAI_X1]    = GetBtFromSgn(maple_controller[dev->bus_id].stick_x);
+		pjs->joy[PJAI_Y1]    = GetBtFromSgn(maple_controller[dev->bus_id].stick_y);
+		pjs->trigger[PJTI_R] = maple_controller[dev->bus_id].trigger_right;
+		pjs->trigger[PJTI_L] = maple_controller[dev->bus_id].trigger_left;
 	}
 	void SetImage(void* img)
 	{
@@ -67,7 +67,7 @@ void mcfg_CreateDevices()
 	#if DC_PLATFORM == DC_PLATFORM_DREAMCAST
 		for(int port = 0; port < MAPLE_NUM_PORTS; port++)
 		{
-			if(port_enabled[port])
+			if(maple_controller[port].enabled)
 			{
 				mcfg_Create(MDT_SegaController, port, 5);
 				mcfg_Create(MDT_SegaVMU,        port, 0);
--- a/core/hw/maple/maple_cfg.cpp
+++ b/core/hw/maple/maple_cfg.cpp
@@ -64,14 +64,19 @@ void mcfg_Create(MapleDeviceType type,u32 bus,u32 port)
 
 void mcfg_CreateDevices()
 {
-#if DC_PLATFORM == DC_PLATFORM_DREAMCAST
-	mcfg_Create(MDT_SegaController,0,5);
-
-	mcfg_Create(MDT_SegaVMU,0,0);
-	mcfg_Create(MDT_SegaVMU,0,1);
-#else
-	mcfg_Create(MDT_NaomiJamma, 0, 5);
-#endif
+	#if DC_PLATFORM == DC_PLATFORM_DREAMCAST
+		for(int port = 0; port < MAPLE_NUM_PORTS; port++)
+		{
+			if(port_enabled[port])
+			{
+				mcfg_Create(MDT_SegaController, port, 5);
+				mcfg_Create(MDT_SegaVMU,        port, 0);
+				mcfg_Create(MDT_SegaVMU,        port, 1);
+			}
+		}
+	#else
+		mcfg_Create(MDT_NaomiJamma, 0, 5);
+	#endif
 }
 
 void mcfg_DestroyDevices()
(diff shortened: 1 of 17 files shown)


Holzhaus:
@skmp What do you think about this?

Holzhaus:
As of 3d5fa60, if both #820 and this PR were merged, multiplayer support should work fine for linux.

skmp:
I'll review and get back to you tomorrow. I don't totally agree with the
way #820 resolves the settings issue, in my view the way forward is less
platforms specific code, not more. But would be great to discuss over it!

On Fri, Aug 28, 2015 at 8:56 PM, Jan Holthuis <notifications@github.com>
wrote:

> As of 3d5fa60
> <https://github.com/reicast/reicast-emulator/commit/3d5fa60ef84069dac5b2a1dc600d5d2a8603d76b>,
> if both #820 <https://github.com/reicast/reicast-emulator/pull/820> and
> this PR were merged, multiplayer support should work fine for linux.
>
> —
> <https://github.com/reicast/reicast-emulator/pull/823#issuecomment-135860117>
> .
>



-- 
~skmp

meveric:
Just a thought:
It makes sense to have all 4 controllers activated even if not all controllers are mapped or actually have controllers connected.
The reason why is simple "more space for savegames" having all 4 controllers activated gives you 8 VMUs to store save data on, even if you only have one controller to play with.
200 Blocks is very little space for some games and it's nice to be able to store savegames on different VMUs even if you only use one controller.

skmp:
It is slightly more complicated, because you can connect "different"
peripherals on the controllers, like vibration and such. But there's
probably a sane default that can be used.

On Mon, Aug 31, 2015 at 4:22 PM, meveric <notifications@github.com> wrote:

> Just a thought:
> It makes sense to have all 4 controllers activated even if not all
> controllers are mapped or actually have controllers connected.
> The reason why is simple "more space for savegames" having all 4
> controllers activated gives you 8 VMUs to store save data on, even if you
> only have one controller to play with.
> 200 Blocks is very little space for some games and it's nice to be able to
> store savegames on different VMUs even if you only use one controller.
>
> —
> <https://github.com/reicast/reicast-emulator/pull/823#issuecomment-136385329>
> .
>



-- 
~skmp

Holzhaus:
Is it possible to connect 8 vmus independent of the number of controllers? This might not be possible on the real dreamcast, but could we do it?

Alternative idea: create a vmu per game. Select the vmu depending on the disk image that was loaded.

skmp:
No the bus locations are on the controllers. You can go wild with
"non-existent" bus designs, but then games go crashing on you. I was
thinking of the savegame idea, but I always found it easier for people to
manage the saves themselves unless there's a really good plan (tm) in
place. Magic is good, but broken magic is far worse than no magic.

On Mon, Aug 31, 2015 at 11:18 PM, Jan Holthuis <notifications@github.com>
wrote:

> Is it possible to connect 8 vmus independent of the number of controllers?
> This might not be possible on the real dreamcast, but could we do it?
>
> Alternative idea: create a vmu per game. Select the vmu depending on the
> disk image that was loaded.
>
> —
> <https://github.com/reicast/reicast-emulator/pull/823#issuecomment-136504665>
> .
>



-- 
~skmp

Holzhaus:
Yeah, but that's a discussion independent of this PR anyway.

core/hw/maple/maple_cfg.cpp
```diff
@@ -68,14 +64,19 @@ void mcfg_Create(MapleDeviceType type,u32 bus,u32 port)
...
 {
-#if DC_PLATFORM == DC_PLATFORM_DREAMCAST
-	mcfg_Create(MDT_SegaController,0,5);
-
-	mcfg_Create(MDT_SegaVMU,0,0);
-	mcfg_Create(MDT_SegaVMU,0,1);
-#else
-	mcfg_Create(MDT_NaomiJamma, 0, 5);
-#endif
+	for(int port = 0; port < MAPLE_NUM_PORTS; port++)
```
skmp: I think this needs to be directly driven from the config, we don't need to have an enabled flag there

core/linux-dist/joystick.cpp
```diff
@@ -9,7 +9,7 @@
 	const u32 joystick_map_axis_usb[JOYSTICK_MAP_SIZE]     = { DC_AXIS_X, DC_AXIS_Y, 0, 0, 0, 0, 0, 0, 0, 0 };
 
 	const u32 joystick_map_btn_xbox360[JOYSTICK_MAP_SIZE]  = { DC_BTN_A, DC_BTN_B, DC_BTN_X, DC_BTN_Y, 0, 0, 0, DC_BTN_START, 0, 0 };
-	const u32 joystick_map_axis_xbox360[JOYSTICK_MAP_SIZE] = { DC_AXIS_X, DC_AXIS_Y, DC_AXIS_LT, 0, 0, DC_AXIS_RT, DC_DPAD_LEFT, DC_DPAD_UP, 0, 0 };
+	const u32 joystick_map_axis_xbox360[JOYSTICK_MAP_SIZE] = { DC_AXIS_X, DC_AXIS_Y, DC_AXIS_LT, 0, 0, DC_AXIS_RT, DC_BTN_DPAD_LEFT, DC_BTN_DPAD_UP, 0, 0 };
```
skmp: DreamcastControllerCodes instead of u32?

skmp:
This looks like a step to the right direction to me. I'd still prefer some more OOP way to go about it (instead of global vars) but this is simple and improves it a lot. Left a few minor comments (not sure they are important enough to guarantee a fix, though). The only sore point is the Active / maple instantiation/cfg issue

core/hw/maple/maple_cfg.cpp
Holzhaus: Call `cfgLoadStr`/Int` here, you mean? I'd rather not do that and rather have a settings object that's decoupled to some degree from the actual config file loading/parsing.

That'd make clean libretro port almost impossible (since that won't use the usual config file and use some keys from `~/.config/retroarch/retroarch-core-options.cfg` instead).  Also, the config file *can* be platform-dependent IMO (like using [`SharedPreferences`](http://developer.android.com/reference/android/content/SharedPreferences.html) on Android for example).

What we could do: Make all the `cfgLoadFoo`/`cfgSaveFoo` a more abstract interface that supports one of these "backends" depending on platform.

Alternative: We could drop the `enabled` flag and introduce an `enable()` method instead, that calls `mcfg_create()` (don't know if that's possible at runtime though).

core/linux-dist/joystick.cpp
Holzhaus: Fixed.

core/hw/maple/maple_cfg.cpp
```diff
@@ -68,14 +64,19 @@ void mcfg_Create(MapleDeviceType type,u32 bus,u32 port)
...
+	{
+		if(maple_controller[port].enabled)
+		{
+			#if DC_PLATFORM == DC_PLATFORM_DREAMCAST
+				mcfg_Create(MDT_SegaController, port, 5);
+				mcfg_Create(MDT_SegaVMU,        port, 0);
+				mcfg_Create(MDT_SegaVMU,        port, 1);
+			#else
+				mcfg_Create(MDT_NaomiJamma,     port, 5);
```
skmp: Naomi only supports a single maple device. It has multiple devices connected over a single jamma i/f

skmp:
Overall looks good. I need to double check after sleeping, too. Can you rebase to latest master?

core/linux-dist/joystick.cpp
Holzhaus: I reverted that commit because `0` isn't a valid member of that Enum and I didn't want to add `DC_NONE` or something like that.

Holzhaus:
> Can you rebase to latest master?

Done.

The crash seems to be caused by the mismatched XML tag in the russian android translations.

baka0815:
I will take a look at this and #843 in the upcoming weeks and try to rework the missing bits for the current codebase.

skmp:
(Cleaning up old tickets - @flyinghead might want to take a look for inspiration)

PR #826, opened 2015-08-29
SDL: Merge sdl/ code into linux-dist/ code

Holzhaus:
This moves the SDL code from `core/sdl/main.cpp` into `core/linux/sdl.cpp` and dumps the removes the duplicates that are already in `code/linux/main.cpp`.

The reason why this doesn't save that many lines is that I did some code formatting fixes while I was at it.

I changed as little functionality as possible, but you might want to test that anyway. (I can't do that right now, since I only have my desktop machine available. It compiles, but I can't see a thing since the proprietary Catalyst drivers don't support EGL/DRI2, so the SDL window stays black).


skmp:
Hmm, I'm not sure I want the sdl code merged w/ the linux, SDL is far more generic (though not used atm). I do like the sdl refactor and the fact that main() is now moved out. Can't we keep that part and have sdl.c/h in core/sdl/ ?

Holzhaus:
Ok. Maybe we can drop X11 in favor of SDL then (and possible also drop the WinAPI stuff?).

Holzhaus:
@skmp 
> Can't we keep that part and have sdl.c/h in core/sdl/ ?

Done.

skmp:
X11 and winapi are quite nice and minimal (and there's very little cost involved in keeping them). While none of them will be "full featured" frontends, there's nothing bad in keeping them from my pov (and, this ensures we don't have a hard dependency on SDL)

Holzhaus:
OK. Should I merge this now?

core/linux-dist/main.cpp
```diff
@@ -23,6 +23,10 @@
 	#include "linux-dist/x11.h"
 #endif
 
+#if defined(USE_SDL)
```
skmp: Should be possible to always just include the header

skmp:
:+1: 
I don't know the actual workings of the SDL code, but I assume you have tested that it works :). @ptitSeb maybe this affects your builds?

ptitSeb:
That will affect my builds indeed. I'll try a new one soon (I built one 1 month ago, using the cmake, and it was working, but so much things have changed)

Holzhaus:
@skmp This doesn't really change the SDL code itself (except for some formatting changes), so I *should* work. It compiles and I can run it, but like I said, I only get a black window since my graphics drivers don't support EGL/DRI2. I also tried getting GL3 working with SDL, but failed miserably :D So I'll just wait for someone to test this properly.

ptitSeb:
I compiled this week end and it worked on the Pandora (but the Makefile is missing the "arm" folder of libpng for neon stuff).

Holzhaus:
I guess we can merge this then?

skmp:
I'll fix master tomorrow (stupid PRI thing broke) and then we merge

On Mon, Sep 7, 2015 at 6:52 PM, Jan Holthuis <notifications@github.com>
wrote:

> I guess we can merge this then?
>
> —
> <https://github.com/reicast/reicast-emulator/pull/826#issuecomment-138339899>
> .
>



-- 
~skmp

Holzhaus:
Ok.

Holzhaus:
@skmp Have you managed to fix it?

skmp:
I think master works alright now. I'll check it out

Issue #630, opened 2015-04-04
Libretro port?


Holzhaus:
Hi,

just wondering, is a libretro port planned?
>Software that targets the libretro API instantly gains access to all of the platforms that have libretro frontends available. That means your game is immediately available to those platforms’ users without worrying about input/audio/video drivers, compatibility issues or the myriad quirks of the underlying host system(s). Likewise, there’s no need to write specialized GUI code for each platform or to use esoteric–and often buggy–cross-platform toolkits, like Qt and GTK+.
>
>Furthermore, once your game supports libretro, all of the libretro frontends’ other features come along for free, such as RetroArch’s real-time rewind, lag-concealing netplay, lossless video recording/streaming and cutting-edge post-processing shader support.
>
>Perhaps best of all: libretro is an open specification that is 100% free to implement, with no licensing fees or hidden strings attached.
>*Source: [http://www.libretro.com/index.php/api/](http://www.libretro.com/index.php/api/)*

MollyM:
I'd love to see this! It'd make setting up my xbox 360 controller a breeze, which currently doesn't work on my raspberry pi 2 with Reicast.

neko68k:
This is a feature request, not a bug. Please only post actual issues. :-)

Holzhaus:
I'm going to looking into implementing this, if I may.

Here are some useful documents:
- [libretro.pdf](https://github.com/libretro/libretro.github.com/blob/master/documents/libretro.pdf)
- [libretro-gl.pdf](https://raw.githubusercontent.com/libretro/libretro.github.com/master/documents/libretro-gl.pdf)

skmp:
Yes, libretro. Our rendering is async, so you can just use it as it is (and
start a background thread for the main sh4 core). @Twinaphex also wanted to
look into this

On Sun, Aug 30, 2015 at 3:02 AM, Jan Holthuis <notifications@github.com>
wrote:

> I'm going to looking into implementing this, if I may.
>
> Here are some useful documents:
>
>    - libretro.pdf
>    <https://github.com/libretro/libretro.github.com/blob/master/documents/libretro.pdf>
>    - libretro-gl.pdf
>    <https://raw.githubusercontent.com/libretro/libretro.github.com/master/documents/libretro-gl.pdf>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/reicast/reicast-emulator/issues/630#issuecomment-136067071>
> .
>



-- 
~skmp

[...]

PR #827, opened 2015-09-02
[WIP] Some inital work for libretro

Holzhaus:
This doesn't contain anything interesting yet, but if you want to, you can leave comments on how stuff should be done :D

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -56,11 +56,15 @@ ifdef FOR_ANDROID
 endif
 
 ifdef FOR_LINUX
-	ifdef USE_SDL
-		RZDCY_MODULES += sdl/
-	else
-		RZDCY_MODULES += linux-dist/
-	endif
+  ifdef LIBRETRO
+    RZDCY_MODULES += libretro/
+  else
+    ifdef USE_SDL
+      RZDCY_MODULES += sdl/
+    else
+      RZDCY_MODULES += linux-dist/
+    endif
+  endif
 endif
 
 RZDCY_FILES := $(foreach dir,$(addprefix $(RZDCY_SRC_DIR)/,$(RZDCY_MODULES)),$(wildcard $(dir)*.cpp))
--- a/core/libretro/libretro.cpp
+++ b/core/libretro/libretro.cpp
@@ -1,0 +1,266 @@
+#include "libretro/libretro.h"
+#include "libretro/reicast.h"
+#include "types.h"
+#include <cstdio>
+#include <cstdarg>
+
+uint32_t video_width =  640;
+uint32_t video_height = 480;
+
+
+// Callbacks
+retro_log_printf_t         log_cb = NULL;
+retro_video_refresh_t      video_cb = NULL;
+retro_input_poll_t         poll_cb = NULL;
+retro_input_state_t        input_cb = NULL;
+retro_audio_sample_batch_t audio_batch_cb = NULL;
+retro_environment_t        environ_cb = NULL;
+
+extern "C"
+{
+	void retro_set_video_refresh(retro_video_refresh_t cb)
+	{
+		video_cb = cb;
+	}
+
+	void retro_set_audio_sample(retro_audio_sample_t cb)
+	{
+		// Nothing to do here
+	}
+
+	void retro_set_audio_sample_batch(retro_audio_sample_batch_t cb)
+	{
+		audio_batch_cb = cb;
+	}
+
+	void retro_set_input_poll(retro_input_poll_t cb)
+	{
+		poll_cb = cb;
+	}
+
+	void retro_set_input_state(retro_input_state_t cb)
+	{
+		input_cb = cb;
+	}
+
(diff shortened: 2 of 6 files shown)


Alcaro:
Some of that looks familiar... as if you copypasted it from Gambatte or something.

>+	bool retro_load_game_special(unsigned game_type, const struct retro_game_info *info, size_t num_info)
>+	{
>+		return false; //TODO (?)
>+	}

Rarely needed. It's fine to leave it empty.

>+	void *retro_get_memory_data(unsigned type)
>+	{
>+		return 0; //TODO
>+	}

You used 0 here, but you used a NULL in dc_init(). Is that intentional?

>-    USE_X11 := 1
>+    #USE_X11 := 1

Let's... let's not break standalone builds, okay? Better to wrap it in an ifdef LIBRETRO or something.

If you've got any other questions, post them here, or join #libretro on Freenode.

Also pinging issue #630.

Holzhaus:
I usually fix the makefiles when stuff is in a mergeable state (so that I can easily recompile during the coding phase...yeah. I'm THAT lazy).

The main roadblock I'm hitting: dc_run loops forever, but libretro calls it per frame (if I understand it correctly).

Alcaro:
That is correct.

Grepping in the source suggests that defining TARGET_BOUNDED_EXECUTION makes it return every 10000 opcodes. If you prefer controlling the returns yourself (you should, otherwise you'll need a weird loop in retro_run()), call sh4_cpu.Stop() at some suitable point (the video callback, for example).

If that doesn't work, we could fiddle with [libco](https://github.com/libretro/bsnes-libretro/tree/libretro/libco) (a coroutine library for C/C++), but it's a pretty ugly mechanism, so try sh4_cpu.Stop() first. Strange how there's no dc_stop() that forwards to that one.

skmp:
The codebase is not meant to be run like that. The TARGET_BOUNDED_EXECUTION toggle will make it in a very "hacky" way, and it's not a stable way to do it yet. Dreamcast is more "modern" than other consoles, and has no fixed video output. Running "one frame" is complicated. Each frame might last a different amount of time, or there might be no new data rendered, or the video might even be "off"

skmp:
For a quick port I'd say we go the threaded way (so run_frame actually waits for a frame, then renders). That's how android and ios work for now. This will make pausing etc broken, which is I guess another issue that needs to be addressed.

The current JIT implementations don't support exiting or pausing, afair.

twinaphex:
Might as well close this.

Holzhaus:
@twinaphex Any plans to merge the existing libretro implementation upstream?

twinaphex:
Nope, all bridges were burnt after I was booted out of this 'team'. And given the lack of willingness to fix crucial bugs in Reicast and priorities being elsewhere, I think it's best if I just stick to my own fork.

twinaphex:
I was the only one in over two years who even sent any commits his way that fixed game compatibility issues anyway instead of mere frontend tinkering, so yeah, he can have fun with the rest of what remains here. I'm not looking back anymore and will do the same as what I did in mupen64plus' case. When upstream stops making sense and has their head stuck between its ass, I view it as a lost cause and just them wasting my time/stifling the progress I could be making. I am starting to view this as another one of these examples.

LoungeKatt:
@twinaphex It is clear you have an issue with the creator of the project and that is your business. There is no need to drag the entire team into it with an egotistical claim that you were the only one to contribute any real work, especially having said that Android, where the work was being done, is part of the project with which you were unfamiliar. Thanks.

twinaphex:
@LoungeKatt Oh I see, some frontend code pusher feels the need to add his final two cents in. 

With all due respect, dude, but I don't need you to add your last two cents in, especially when you don't even know what you are talking about.

RetroArch Android has existed for longer than you have probably been coding. Please keep your mouth shut if you don't know what you are talking about.

And with that said, I am out. I will decide for myself exactly who I drag into what By all means self-identify with this little project all you want and the progress (or lack thereof) it has made for years, but you will no longer be seeing any pull requests from me unless @skmp starts humbling himself and actually starts giving a shit again about improving core emulation. Or hell, at least fix the regressions you caused in nullDC and at least bring it up to nullDC's level for god's sake. That is better than leaving it perpetually broken for years and playing silly games with your endusers.

LoungeKatt:
@twinaphex "As for Reicast for Android, I'll try looking into it. I just never got to it before." (#947).
Rather than taking a polite request not to be so obnoxious, you decided to run your mouth again. I have little concern about your opinion of my contribution to this project. The reason is not the progress in RetroArch or your personal contribution to this project. The reason is that the best you have to say is some insulting, egotistical trash about being some sort of messiah. You, as a person, do not fit in here because you, as a person, are not good enough. It has little to do with code.

LoungeKatt:
@Holzhaus There are no plans, but all comments by the author have been removed to avoid having the repository dragged down by immature, offensive, and egotistical ramblings. 

This post is only meant for archival purposes.

Issue #831, opened 2015-09-11
Add some user interface to Windows build [$130]


ofry:
For sample, Open file, Open BIOS, Select BIOS, Save/load state, etc.

Holzhaus:
We should probably separate the Desktop GUI from the emulator code and use an approach similar to what mupen64plus does: http://m64py.sourceforge.net/

That way we'd have a unified code path for all platforms and don't introduce big dependencies in case we don't need a GUI.

ofry:
Yes, it's possible too ^^

ofry:
#630 I think it's related. Because if we port it to libretro, it can be good GUI for this emulator :+1:

skmp:
I'm always in more "integrated" things than just a front end (So that you can actually pause and change settings during run time). Most of the user friendly things can't be done yet though (save states don't exist, input is only keyboard for windows)

Holzhaus:
That doesn't conflict with the python-gui approach. You can actually call c functions from python and vicd versa. I was already looking into doing something like that, but stumbled over a bug in my SDL2. I see if I can work around that.

skmp:
Well, that could work then. (Keep in mind python is a very foreign and pita to setup thing for windows users, though)

Holzhaus:
Well, there are things like [pyexe](http://www.py2exe.org/) which turn python programs into exe files that are runnable without python installed. The m64py people even have a Windows portable build for download, so has to be a way to make it user friendly. I'll look at the code and get in touch with them eventually.

We should think about separating the current platform-specific user-interface stuff (loading settings, creating a render surface, and sorts) from the core reicast code, so that we can turn the latter into a shared library that both the current UI and the future GUI can access.
That way, a user can simply install reicast and if he/she later decides that a GUI is required, it can be installed afterwards, but you don't have to compile and keep a second copy of the core reicast stuff. This mainly affects linux - windows users usually expect a GUI (and the display server is installed anyway), so if we had a GUI, I think we should always include it in the windows version.

Holzhaus:
I've got a basic Python GUI working now, which does not integrate with reicast just yet, but the concept seems to work. After the GUI is created, I get the HWND/XWindowID of one of the widgets and pass it to some C code that creates an SDL2 window with OpenGL 3.1 support from it and draws a black background on it. Works fine as far as i can see.

[...]

PR #832, opened 2015-09-12
Move to SDL2 for OpenGL 3.1 and GLES support

Holzhaus:
~~Unfortunately, I needed to move to SDL2 to make this work. I'm not really happy with introducing yet another code path, so we should decide if we want to keep SDL1.2 support (without OpenGL3) or if we wan't to move to SDL2.~~

**EDIT:** What this PR does:
- Migrate from SDL1.2 to SDL2 and drop SDL1.2 support
- We can now initialize the OpenGL 3.1 context in a platform-independent way via SDL2 (SDL1.2 did not support OpenGL 3.1)
- OpenGL ES is also initialized via SDL2 instead of using custom EGL code (it's still there for Android though, but maybe we can also use SDL2 on Android?).

As a Bonus, I can now run reicast with OpenGL ES on my Linux desktop (the egl code always failed, but via SDL2 it works fine).

### DIFF
--- a/core/khronos/GL3/gl3w.c
+++ b/core/khronos/GL3/gl3w.c
@@ -1,6 +1,25 @@
 #include <GL3/gl3w.h>
 
-#ifdef _WIN32
+#if defined(USE_SDL2)
+	#include <SDL2/SDL.h>
+	static void open_libgl(void)
+	{
+		SDL_GL_LoadLibrary(NULL);
+	}
+
+	static void close_libgl(void)
+	{
+		SDL_GL_UnloadLibrary();
+	}
+
+	static void *get_proc(const char *proc)
+	{
+		void *res = NULL;
+		res = (void*)SDL_GL_GetProcAddress(proc);
+		return res;
+	}
+
+#elif defined(_WIN32)
 #define WIN32_LEAN_AND_MEAN 1
 #include <windows.h>
 
--- a/core/khronos/GL3/gl3w.c
+++ b/core/khronos/GL3/gl3w.c
@@ -1,6 +1,6 @@
 #include <GL3/gl3w.h>
 
-#if defined(USE_SDL2)
+#if defined(USE_SDL)
 	#include <SDL2/SDL.h>
 	static void open_libgl(void)
 	{
--- a/core/rend/gles/gles.cpp
+++ b/core/rend/gles/gles.cpp
@@ -355,7 +355,7 @@ int screen_width;
 int screen_height;
 
 #if (HOST_OS != OS_DARWIN) && !defined(TARGET_NACL32)
-#ifdef GLES
+#if defined(GLES) && !defined(USE_SDL)
 	// Create a basic GLES context
 	bool gl_init(void* wind, void* disp)
 	{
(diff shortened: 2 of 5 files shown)


Holzhaus:
@skmp I wonder if we can drop SDL1.2 in favour of SDL2?

In the long term, we should try to make SDL2 the default for all platforms (that doesn't mean we need to drop the non-SDL implementations, just that SDL2 becomes the default build option, so that we have *one* codepath that we can concentrate on.

Holzhaus:
I rebased to make wercker happy.

Currently, it seems that the only plaform actually using SDL is Pandora. SDL2 seems to have [pandora support](http://hg.libsdl.org/SDL/file/d3ad01d50bf0/docs/README-pandora.md), so it would be cool if somebody could test this.

ptitSeb:
What are you doing to ma poor SDL port...

Ok, I'll test SDL2 (it should work, but on the Pandora, SDL1.2 is enough, SDL2 brings nothing significants)

Holzhaus:
Yeah, but SDL2 has the benefit of supporting GL 3.1 and GLES on desktops. Also, SDL2 is the way to go when tackling #831. And if SDL2 works fine on Pandora, we don't have to maintain the extra code path for SDL1.2.

ptitSeb:
Yeah I understand, no issues.

skmp:
@ptitSeb so, does this work alright w/ pandora?

ptitSeb:
I'll test that.

ptitSeb:
It compile mostly fine, but I have trouble creating the Window (I'm not even at the context creation). It may be my hacked version of SDL2 (dual GL / GLES2 support is tricky). I need to debug the SDL_CreateWindow call first...

ptitSeb:
Ok, issue was on my side. It's ok for me, SDL2 seems to works enough (I haven't checked the nubs yet, but it should be ok).

skmp:
Alright. I'm only a litl bit worried about distros that don't have SDL2 (ubuntu 12.04 LTS among others, possibly rpi2 distros). I guess i'm too legacy-paranoid, right?

Holzhaus:
> I guess i'm too legacy-paranoid, right?

No, it's a valid concern. But Linux users need to know how to compile software on their own anyway (since Reicast itself has to be compiled from source), so doing `./configure`/`make`/`make install` for SDL2 doesn't make that much of a difference.

ptitSeb:
For Rpi2, the GLES context creation is specific. Without proprer support, a simple `./configure && make ` will not works. It seems there is a `raspberry` folder in the video driver of SDL2, so I guess some support has been implemented.
Other platform, like ODroid, have SDL2 support.

Holzhaus:
From the [SDL wiki](https://wiki.libsdl.org/Installation#Raspberry_Pi):
> If you want to build SDL on your Raspberry Pi directly, just build it "the Unix way," as the RPi is, more or less, a standard Linux system. It can be built with X11 support, and also can do OpenGL ES rendering directly to the screen without an X server, to save resources. In theory, Wayland could also be supported, but at the time of this writing it doesn't configure/compile correctly (send patches!).

Holzhaus:
Merge?

ptitSeb:
On my side, I'm ok with the merge (I'll probably have to work on the Pandora joystick  code, but it's not a blocker issue).

dankcushions:
:+1:

skmp:
Doit!

robertybob:
@Holzhaus :+1:

PR #836, opened 2015-09-16
Update wercker.yml

skmp:
For some reason looks like the install-packages script is broken...

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -1,4 +1,4 @@
-box: ubuntu
+box: ubuntu:15.04
 
 build:
     steps:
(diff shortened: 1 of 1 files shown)


skmp:
This looks like some issue w/ the docker base image to me

skmp:
Hah, nice idea... but build still failed :p

skmp:
That last one seems to have nailed it. Interestingly, my similar attempt (update & co, but still using the pre-defined install deps step) failed. Hum. Wheeeee I guess :D

Holzhaus:
Maybe it was the `clean`? I suppose that whole error was due to some crazy caching issue.

skmp:
Can you make a clean PR just for it and merge it?

Holzhaus:
Done.

PR #837, opened 2015-09-17
sh4/mmu: C++ exception based sh4 exception support. Won't run wince yet.

skmp:
- Rewrite mem ops to only modify regs after exception path
- Throw & catch logic for interpreter that raises the exception
- Re-enabled some commented mmu code

DC Linux does work though. No FB emulation but we get to the console prompt :3
mmu-wip-2015-09-17 21_23_27-reicast - microsoft visual studio administrator

### DIFF
--- a/core/build.h
+++ b/core/build.h
@@ -115,7 +115,7 @@
 
 */
 
-//#define NO_MMU
+#define NO_MMU
 
 #define DC_PLATFORM_MASK        7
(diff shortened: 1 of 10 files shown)


skmp:
Hrfm, I fixed nullDC to work on vs 2013 and can't get wince stuff working on it either. Not even bleemcast. I really wonder, which version *did* work? -.-

Issue #427, opened 2014-03-28
Reicast on Linux


moocow1452:
So, as far as other ports outside of the mobile space, where would be a place to start if we want to run it on Linux? Would a dedicated Raspberry Pi or Beaglebone Black build be ideal, or can a proper x86 build be made?

marnold:
hello 

I tried with latest commit and have this gdb trace 

Reading symbols from ./reicast.elf...done.
(gdb) r
Starting program: /home/matt/src/reicast-emulator/shell/lin86/reicast.elf 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Home dir is: /home/matt/.reicast/
Linux paging: 00001000 00001000 00000FFF
evdev open: No such file or directory
SDK: Found 'HuiJia  USB GamePad' joystick with 8 axis and 16 buttons
MAP 00800000 w/ 25165824
MAP 20000000 w/ 25165824
MAP 04000000 w/ 16777216
MAP 06000000 w/ 16777216
MAP 0C000000 w/ 0
A8
vmem reserve: base: D7510000, aram: f7510000, vram: DB510000, ram: E3510000
Resetting mem
Mem alloc successful![New Thread 0xd54cfb40 (LWP 6691)]

libwebsockets test server - (C) Copyright 2010-2013 Andy Green <andy@warmcat.com> - licensed under LGPL2.1
Initial logging level 7
Library version: 1.3 c11b847
IPV6 not compiled in
libev support not compiled in
 static allocation: 4452 + (12 x 65536 fds) = 790884 bytes
 canonical_hostname = magmar-1
 per-conn mem: 116 + 1554 headers + protocol rx buf
 Listening on port 5678
Loaded /home/matt/.reicast/data/dc_boot.bin as bootrom

Loaded /home/matt/.reicast/data/dc_flash_wb.bin as nvram

Using Interpreter
[New Thread 0xd19efb40 (LWP 6692)]

GDI : -13132 tracks

Program received signal SIGSEGV, Segmentation fault.
0xf79c9a6f in _IO_vfscanf () from /lib/i386-linux-gnu/i686/cmov/libc.so.6
(gdb) c
Continuing.
Fatal error : BM_LockedWrite and NO REC
 in BM_LockedWrite -> ../../core/hw/mem/_vmem.cpp : 622 


Program received signal SIGTRAP, Trace/breakpoint trap.
[...]

PR #839, opened 2015-09-22
Linux/man: Use gdi as example image file

Holzhaus:
Resolves https://github.com/reicast/reicast-emulator/issues/427#issuecomment-138111981.


skmp:
:+1: 

Meanwhile, and idea on how to fix wercker?

PR #841, opened 2015-09-28
apple: Add primitive core audio support

skmp:
This needs to be lock correctly and buffer correctly before making it to master. But it's a place to start!

Sound does work on OS X, on iOS I get silence (but no crashes). iOS uses 1024 frame blocks and that could be one of the issues there

### DIFF
--- a/core/oslib/audiobackend_coreaudio.cpp
+++ b/core/oslib/audiobackend_coreaudio.cpp
@@ -1,0 +1,142 @@
+/*
+    Simple Core Audio backend for osx (and maybe ios?)
+    Based off various audio core samples and dolphin's code
+ 
+    This is part of the Reicast project, please consult the
+    LICENSE file for licensing & related information
+ 
+    This could do with some locking logic to avoid
+    race conditions, and some variable length buffer
+    logic to support chunk sizes other than 512 bytes
+ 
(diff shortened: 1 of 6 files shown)


skmp:
@Holzhaus do you have any ideas on how to fix wrecker? it's really becoming a PITA :<

core/oslib/audiobackend_coreaudio.cpp
```diff
@@ -0,0 +1,142 @@
...
+ */
+
+#include "oslib/audiobackend_coreaudio.h"
+
+#if HOST_OS == OS_DARWIN
+
+//#include <CoreAudio/CoreAudio.h>
+#include <AudioUnit/AudioUnit.h>
+
+AudioUnit audioUnit;
```
Holzhaus: Shouldn't these be static?

skmp: Maybe, but I generally don't use the "static" keyword cus I don't like how it sounds/looks/feels. Linkage control has nothing to do with "staticity". So I add it only when it becomes a problem...

mk8itra1n:
Looks like you got sound to work in reicast. Amazing progress already only days into returning to reicast development. Keep up the good work @skmp

PR #869, opened 2015-11-14
[Android] Update values-ru/strings.xml

6lackmag3:
(values-ru) added missing strings, reworked translation


skmp:
Double checked via google translate and these look a major improvement over the previous ones. Thanks!

PR #873, opened 2015-11-29
reicast-joyconfig.py: store btn_escape in the right section

gizmo98:


skmp:
@Holzhaus can you verify this one? I don't know how the python stuff works. @gizmo98 as this is your first contribution can ya read #455 please ? ~ off to look into the other pending stuff ~

gizmo98:
@skmp according to evdev.cpp btn_escape is located in section "emulator"
https://github.com/reicast/reicast-emulator/blob/915d6b26d67ebda19e6029eb3547b6c6295cb7ef/core/linux-dist/evdev.cpp#L170

Holzhaus:
Yeah, makes sense. It's just there for quitting the emulator (the actual dreamcast doesn't have an "Escape" button), hence it belongs into the emulator section.

PR #890, opened 2015-12-26
Fix for keyrepeat issue

7oxicshadow:
This probably is not noticeable by most Linux users as I suspect that everyone is using a controller with the emulator but there is an annoyance with Linux and the way it handles keyboard events.

By default it assumes that if someone is holding a key down the user is requesting a rapid key cycle so it "pulses" the key request. In Reicast this results in irratic movement in games (such as pulsed acceleration in racing games) and it also fills the keyboard buffer so that when the key is released it adds a very noticable delay before the emulator responds.

The fix in this pull request turns the key repeat feature off when the X11 window is created.


LoungeKatt:
@7oxicshadow was it just the one commit for actual the fix?
I was trying to rebase it to clean up the PR a bit lol
https://github.com/reicast/reicast-emulator/tree/fix-key-repeat

7oxicshadow:
Yeh, its just the one commit in x11.cpp.

I have no idea how i screwed this one up but it went nuclear

LoungeKatt:
git doesn't always use commons sense. It probably tried to fast forward and used it's more is better logic.

AbandonedCart:
This should definitely be reopened with the commit added to a fresh branch.
The commit is https://github.com/reicast/reicast-emulator/pull/890/commits/28add301fa35caeaaafcb44b8edb5b6948c485d9

PR #902, opened 2016-01-30
Fix the check for the enviroment variable

### LINKED ISSUE #847: Crash on emulator launch
Running reicast built from commit 915d6b26d67ebda19e6029eb3547b6c6295cb7ef on Linux
Attempting to launch reicast will immediately crash with this message
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_S_construct null not valid
Aborted (core dumped)

I tried compiling older releases to see if they work but the builds kept failing.

philenotfound:
I tried running reicast and it immediately died with:

terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_S_construct null not valid
Aborted

I was missing the environment variable XDG_CONFIG_DIRS

Turns out find_system_config_dirs() checked for the existence of
XDG_DATA_DIRS but then tried to use XDG_CONFIG_DIRS (non-existent in my
case).

fixes #847

### DIFF
--- a/core/linux-dist/main.cpp
+++ b/core/linux-dist/main.cpp
@@ -347,7 +347,7 @@ string find_user_data_dir()
 std::vector<string> find_system_config_dirs()
 {
 	std::vector<string> dirs;
-	if (getenv("XDG_DATA_DIRS") != NULL)
+	if (getenv("XDG_CONFIG_DIRS") != NULL)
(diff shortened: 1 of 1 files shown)


Holzhaus:
Thanks for the fix! Must have overlooked that.

PR #913, opened 2016-04-11
Add an option to core.mk to build android without webui if wanted

pvezien:
Remove the line "WEBUI := 1" from Android.mk to test.


AbandonedCart:
This was labeled a test from 2016 and the account has since been deleted with an unsigned CLA. Please reopen if further progress is made.

PR #917, opened 2016-04-22
Allow Linux/Windows build to close cleanly

7oxicshadow:
*Add the ability to stop the render thread
*Add the ability to stop the ngen thread
*Add the ability to detect the window close event
*Add gl_context_destroy function
*Close any open controllers
*Allows the nvmem to save

### DIFF
--- a/core/cfg/ini.cpp
+++ b/core/cfg/ini.cpp
@@ -166,25 +166,49 @@ bool ConfigFile::get_bool(string section_name, string entry_name, bool default_v
 
 void ConfigFile::set(string section_name, string entry_name, string value, bool is_virtual)
 {
-	ConfigSection* section = this->get_section(section_name, is_virtual);
+	bool local_virtual;
+
+	#ifdef NO_VIRTUAL_CFG
+		local_virtual = false;
+	#else
+		local_virtual = is_virtual;
+	#endif
+
+	ConfigSection* section = this->get_section(section_name, local_virtual);
 	if(section == NULL)
 	{
-		section = this->add_section(section_name, is_virtual);
+		section = this->add_section(section_name, local_virtual);
 	}
 	section->set(entry_name, value);
 };
 
 void ConfigFile::set_int(string section_name, string entry_name, int value, bool is_virtual)
 {
+	bool local_virtual;
+
+	#ifdef NO_VIRTUAL_CFG
+		local_virtual = false;
+	#else
+		local_virtual = is_virtual;
+	#endif
+
 	std::stringstream str_value;
 	str_value << value;
-	this->set(section_name, entry_name, str_value.str(), is_virtual);
+	this->set(section_name, entry_name, str_value.str(), local_virtual);
 }
 
 void ConfigFile::set_bool(string section_name, string entry_name, bool value, bool is_virtual)
 {
+	bool local_virtual;
+
+	#ifdef NO_VIRTUAL_CFG
+		local_virtual = false;
+	#else
+		local_virtual = is_virtual;
+	#endif
+
 	string str_value = (value ? "yes" : "no");
-	this->set(section_name, entry_name, str_value, is_virtual);
+	this->set(section_name, entry_name, str_value, local_virtual);
 }
 
 void ConfigFile::parse(FILE* file)
--- a/core/core.mk
+++ b/core/core.mk
@@ -26,19 +26,19 @@ ifndef NO_REC
 endif
 
 ifndef NOT_ARM
-    RZDCY_MODULES += rec-ARM/
+    RZDCY_MODULES += rec-ARM/ rec-common/
 endif
 
 ifdef X86_REC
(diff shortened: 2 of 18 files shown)


substring:
hey @7oxicshadow 
been trying your patch. I'm compiling reicast for raspberry/linux, no x11, the link crashes with unknown references to `x11_gl_context_destroy()` in `core/linux-dist/main.cpp:469` as well as for `x11_window_destroy()` in `core/linux-dist/main.cpp:472`
I guess you are missing some checks for the compiling/running environment

7oxicshadow:
Thanks for letting me know. I will update it when I can.

In the mean time could you try manually editing your main.c and adjusting the code at line 468:

#if defined(SUPPORT_X11)
 	/*Close the GL context */ 
 	x11_gl_context_destroy(); 
 	 
 	/* Destroy the window */ 
 	x11_window_destroy(); 
#endif

substring:
hey thanks @7oxicshadow 
Compiled fine, but won't set time on my build after following http://wiki.reicast.com/index.php/Help#How_to_set_the_time

7oxicshadow:
@substring Thanks for testing.

On an X11 build we detect a request for a program to close via the "ESC" key press or the window being closed by clicking the "X" button on the window itself.

I will have to look if there is a different path for exiting the program on a non X11 build. Have you been testing on a raspberry pi? how do you close the program?

substring:
this is what the log says on exit :
```
Fatal error : death by escape key
in input_evdev_handle -> ../../core/linux-dist/evdev.cpp : 333
```
I'm testing on a pi3, yes. I mapped a button on my pad to exit (btn_escape). I guess the routine is not the same if reicast is exited that way

7oxicshadow:
Ah I see now.
if you edit evdev.cpp in the following way it should work.......(in theory)

Goto line 300 and add 
```
	void ngen_terminate();  
	void rend_terminate();  
```
Goto line 333 and replace "die("death by escape key");" :
```
	/* Terminate the main program loop */  
	rend_terminate();  
	ngen_terminate();  
```
If this does not work I will test it when I get home from work.

substring:
nope, it can't find `ngen_terminate()` at link. Looking at your diff, looks like the function defined in `core/rec-x64/rec_x64.cpp` which is an architecture specific file. I'm building for arm. I guess such code shouldn't be arch specific.

7oxicshadow:
@substring 

This issue is a little more involved as it appears to use 'asm' for the arm stuff. I will dig deeper at the weekend.

If saving the nvram is the main issue you are having you could make a quick temporary change to evdev.cpp

1. Remove any previous changes to evdev.cpp
2. goto line 300 and add:
```
void dc_term();
```
3. goto line 333 and change it to the following:
```
dc_term();
die("death by escape key");
```
This should save the eeprom before force closing the app and stop the annoying date problem.

substring:
@7oxicshadow I guess something is wrong somewhere on my side. The config option `Dreamcast.RTC` is set, but I don't get why the DC time is not correct ... nevermind, will have a look later. Thank you for your help :)

rcampbel3:
it would be really nice to get these linux fixes merged. Can we get someone with write access to review this?

LoungeKatt:
Looking into it now.

BenMcLean:
Any update on this? :)

substring:
@7oxicshadow confirming it's working. Nice job on this one :)

substring:
@7oxicshadow have you tried building for linux x86 ? It fails with:
```
obj-dreamcast-x86/linux-dist/main.build_obj: In function `start_shutdown()':
/work/output/build/reicast-ca837e19b8c8247f8ff9efd699ff49533e834d9a/shell/linux/../../core/linux-dist/main.cpp:440: undefined reference to `ngen_terminate()'
```
Builds fine for x86_64

skmp:
@7oxicshadow I'll try to take a look over this weekend. Do you mind signing the CLA?

7oxicshadow:
I have signed the CLA. It has been a while since I looked at this, but I think going forward it would be best to consider creating splitting the emu.cfg into 2 config files. An RTC config and an options config.

Currently, any front end using the -config command line parameter will have to use my build time parameter NO_VIRTUAL_CFG if they want to save the RTC. By having a dedicated RTC.cfg file you could have a working RTC and virtual config support as you originally intended [Just make sure that the RTC.cfg isn't opened as virtual :) ]

skmp:
There was a fix for the virtual cfg, now only the set keys become virtual. Does that address that issue?

skmp:
Looking closer at the diff, I think we should re-work this in a way that is less linux specific. These issues also happen in android, to some extent, and windows, right?

LoungeKatt:
I think some of the integration checks got frozen. No way it took 9 hours to run a compile.

sl1pkn07:
seems need rebase, i can't apply this pr into master

AbandonedCart:
Closing this due to lack of activity. Please reopen if it is resumed.

Issue #915, opened 2016-04-16
Cross compiling reicast for Raspberry Pi with GCC 5.30


substring:
Hi,

Just to mention I had to add -D_GLIBCXX_USE_CXX11_ABI=0 to avoid some problems with std::string when compiling with GCC5, otherwise it couldn't read correctly a evdev_mapping_N from emu.cfg, whichever value you set

substring:
thanks @gizmo98 for making the PR ;)
no need to keep the issue open then

PR #924, opened 2016-05-08
Makefile: Fix gcc 5.x std::string issue

gizmo98:
Evdev mapping file cannot be read if GCC5 is used.
https://github.com/reicast/reicast-emulator/issues/915


skmp:
I'd rather fix the source rather than working around compiler options. Is that possible?

sabula310:
when will you guy updated reicast it been a year already

skmp:
(Closing this, will take a look on the underlying issue if it pops up again)

PR #926, opened 2016-05-13
ftrc fix - revert back to nulldc defaults

twinaphex:


twinaphex:
Confirmed to fix audio clipping issues with samples in Dead or Alive 2 and Sonic Adventure 1 in the libretro port. Should be good to go.

sabula310:
that cool to hear that fix audio issues in dead or alive 2 but can you do
star wars racer when you start to race its makes this sound over and over
thank for your time

PR #927, opened 2016-05-13
Delete spurious code - fixes sprite palette issue with games like

twinaphex:
Street Fighter III 3rd Strike/Capcom Vs SNk 2, possibly others


skmp:
Huh, that was retarded. Rather unsuprising to see these kind of bugs though :<.

skmp:
texcache: fix vram_lock to actually lock the entire texture

PR #930, opened 2016-05-16
Nicer multiplayer support

substring:
This one doesn't depend on an ini option, but rather relies on the number of evdev_device_id_X > -1
If USE_EVDEV is not set, fallback to the previous way of doing : just 1 controller


skmp:
@substring what is the goal for this MR? It feels like this code shouldn't be in the maple_cfg part, but in the linux shell files?

substring:
Hey !

I thought it was unnecessary to add a new value in the .cfg to specify the
number of players, but simply rely on the number of evdev devices declared.

baka0815:
Superseded by #1288 / #1308.

PR #877, opened 2015-12-08
X360 controller support for Windows + Rumble

KrossX:
Some basic controller input for Windows, but rumble required a PuruPuru Pack so added that too.

Quite basic, no error check neither fancy continuous rumble mode but it works. Rez feels nice~

PR #964, opened 2016-09-13
Linux rumble support

Holzhaus:
This is a follow-up PR to #877.


skmp:
@Holzhaus I merged #877 before this one. Shal I merge on top?

baka0815:
This includes the necessary parts for the Linux support of #877. As far as I can see the changes are the same as the ones mentioned by @Holzhaus in the #877 which I have verified and tested to be working.

So, for what it's worth: LGTM

skmp:
*crosses fingers*

PR #972, opened 2016-09-23
linux-dist/evdev: Use std::Map.find() instead of operator[]

Holzhaus:
Since EvdevControllerMapping does not have a default constructor,
we can't use the operator[];

    data_type& operator[](const key_type& k) - Returns a reference to
    the object that is associated with a particular key. If the map
    does not already contain such an object, operator[] inserts the
    default object data_type().

Instead, we use the std::map.find() function.

### DIFF
--- a/core/linux-dist/evdev.cpp
+++ b/core/linux-dist/evdev.cpp
@@ -285,7 +285,7 @@
 						return -3;
 					}
 				}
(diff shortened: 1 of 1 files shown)


Holzhaus:
Since this just a one-liner and shouldn't break anything, I'll merge this now.

PR #976, opened 2016-09-24
These files do not need to be executable

orbea:
A lot of files which are executable (`0755`) so I changed them to be `0644`.


orbea:
Given the lack of interest I will just close this...

Issue #977, opened 2016-09-24
rec_v3 design plan


skmp:
Long term idea: Front end -> IL -> HLR -> LIR -> NIR -> binary

First pass
- linked list of ops
- full load_reg/store_reg
- no opts
- shil_params -> add ref
- reg assignment-> implement for refs

Goals:
- Cleaner code
- Documented code
- Less custom crufty stuff
- (Full) register allocation for vectors and temps
- Perform at least as well as the current backends
- Provide flexibility for future work, like shop_mem_transl, explicit values and multi-block
- After first iterations to gauge things, explore other projects for inspiration (LLVM/Xenia/etc)

skmp:
(closing this as it is kinda dead)

PR #978, opened 2016-09-26
WIP: ideas for rec_v3

skmp:
WIP for #977. 

I tried quickly adapting the existing IR w/ the introduction of virtual regs as a step towards SSA-form IL. Once register assignment is implemented (need to compute live spans) it should work. 

Once this is running, the next step is to re-structure the IR to correctly handle values, then implement LSE & register allocation and update the backends.


skmp:
(this is outdated)

PR #985, opened 2016-10-14
fixing x11 keyboard support

coadde:
Add analog stick, L/R analog triggers and Y/Z/C/D buttons

### DIFF
--- a/core/linux-dist/x11.cpp
+++ b/core/linux-dist/x11.cpp
@@ -3,6 +3,7 @@
 #include <X11/Xlib.h>
 #include <X11/Xatom.h>
 #include <X11/Xutil.h>
+#include <X11/XKBlib.h>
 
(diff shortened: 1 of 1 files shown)


elBlanquito:
Per, http://sega.wikia.com/wiki/VMU, VMU was used in Marvel Vs. Capcom 2. Where are you with respect signing the license Coade? Did you have ethical objections or just have this fall off your radar?

skmp:
True, but someone has to open a new MR. It's much better for peace of mind if new commits of "unsigned" origin do not make it in. We can try to send the author an email?

LoungeKatt:
Not sure if you took that as anyone could resubmit the specific commit, but it literally moves two lines outside a } to inside it. Cherry-picking that is reason enough to get banned from github.

PR #986, opened 2016-10-19
alsa-backend: initialize variable "sub unit dir"

verybadsoldier:
Really minimal change to fix alsa initialization in some environments:

This variable is passed into alsa as a pointer. Alsa will write to it but also read (!) from it. So the read will give undesired behavior because variable was not initialized. Depending on current alsa-config/hardware it leads to an error "Invalid argument" when submitting the parameters by invoking "snd_pcm_hw_params" a few lines later.

Initializing the variable to -1 fixes it and means "set to exact value or less" (http://alsa.opensrc.org/HowTo_Asynchronous_Playback)

### DIFF
--- a/core/oslib/audiobackend_alsa.cpp
+++ b/core/oslib/audiobackend_alsa.cpp
@@ -13,7 +13,7 @@ static void alsa_init()
 
 	snd_pcm_hw_params_t *params;
 	unsigned int val;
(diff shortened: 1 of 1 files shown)


skmp:
Good catch. Have you read #455 (Code should be MIT-licensed for new controls)?

PR #1002, opened 2017-01-29
Improve keyboard input handling on X11

alexisrozhkov:
* Add support for triggers and analog stick

* Enable loading keyboard mapping by specifying cfg file for evdev_mapping_... in input section of emu.cfg instead of hardcoding it.

* Fix dangling pointer in cfg file loading


baka0815:
This has been fixed in other ways by now.

PR #1019, opened 2017-05-07
Updates for newer APIs / remove outdated code

LoungeKatt:


LoungeKatt:
Bleh. It needs to compile against legacy HTTP, but not include it.

PR #1060, opened 2017-11-13
Readme: Add slack link

skmp:
@tolis764 @LoungeKatt FYI


skmp:
Also, @Holzhaus if you're still around :)

PR #1086, opened 2018-03-22
Travis CI fix: install ant

skmp:


skmp:
Looks like Travis does not come with ant installed by default now? Manually installing ant works around the problem. One more reason to switch over to gradle I guess

PR #1087, opened 2018-03-22
Basic snapcraft configuration

skmp:


skmp:
After further investigation it looks like snapcraft doesn't really cater well for GUI applications right now. I guess it would make more sense to setup a couple of official build/package scripts for ubuntu, arch, etc for now, and we can revisit when things change.

PR #1092, opened 2018-03-25
fix gdi track parsing in unixx

i30817:
Same problem as https://github.com/libretro/reicast-emulator/pull/24

in unix that iteration would not work if the path of the original .gdi file (where we are finding the 'parent directory') was:

1. not existent, ie a gdi on the same directory as the current. It was iterating until the index 3 and stop there. Disaster strikes later ofc, if the first characters weren't C:\ or other thing like that.
2. in a relative subdirectory that is 'small enough' ie: a/crazy-taxi.gdi. In this case the last '/' wouldn't be found, the derived string would be 'crazy-taxi.gdi' and the file not found ofc.

Anyway, this can be solved simply by searching the whole string and using a signed integer. It will go up to -1, stop iteration and increase to 0 on the len++.

Dunno if using relative subdirectories for the tracks on *other* platforms works because of the path separator being different from what's inside the gdi but with this code fixed it works on linux with / at least.

### DIFF
--- a/core/imgread/gdi.cpp
+++ b/core/imgread/gdi.cpp
@@ -36,9 +36,9 @@ Disc* load_gdi(const char* file)
 	strncpy(path, file, sizeof(path));
 	path[sizeof(path) - 1] = '\0';
 
-	size_t len = strlen(path);
+	ssize_t len = strlen(path);
 
-	while (len>2)
+	while (len>=0)
 	{
 		if (path[len]=='\\' || path[len]=='/')
 			break;


LoungeKatt:
Windows wouldn't accept `ssize_t` natively, but setting a bounds of 0 should still be respected on all platforms. If it is still an issue using `size_t`, then the entire variable can be added as part of the conditional.

i30817:
honestly, then make the _WIN32 ifdef apply to the declaration of the type. [Something like.](https://stackoverflow.com/a/35368387/214260)

The 'corrected' code is just reintroducing the bugs though i suppose i should have known windows would break posix typedefs.

LoungeKatt:
I'm not sure how `size_t` would reintroduce the bug. As explained in https://stackoverflow.com/a/15739705, ssize_t is when you require a -1 error case at the cost of half the possible positive length, but you don't actually use -1. It actually just cuts the possible size in half to say that there is a value lower than 0 available.

i30817:
I pretty much explained that `while pos > 2` is wrong already and showed cases where it completely explodes. So if you want to use size_t the loop has to change.

```
        size_t len = strlen(path);
	while (true)
 	{
 		if (path[len]=='\\' || path[len]=='/' || len == 0)
 			break;
 		len--;
 	}
        if (path[len]=='\\' || path[len]=='/')
             len++;
```
should work everywhere even if it looks uglier.

I'd personally prefer 
```
        int len = (int) strlen(path);
	while (len>=0)
 	{
 		if (path[len]=='\\' || path[len]=='/')
 			break;
 		len--;
 	}
len++;
```
but that's just me. I wonder if i have to fix the RA commit too.

LoungeKatt:
There is no disagreement that on a system without drive designation, that the lost path components are an issue. I also cannot confirm without testing that it won't also cause issues on windows with a drive designation (hence simply excluding windows to move the fix along).

I'm not sure if it is any better of an explanation, but there is also https://stackoverflow.com/a/22267668. 
The simplest way to word it would be that ssize_t will chop size_t in half to get an error code that, in your case, is never used.

i30817:
It's not rocket science. If you insist on using `while (pos < 2)` the code is wrong and several valid paths become invalid. I just proposed another PR to the the RA reicast fork that makes use of the first alternative variant on the my last post, that should work all platforms using \\ or / as separators.

I'll also remind you that there are more problems to solve in this code if you're touching it. Making sure that the tracknames directory separators (if any) on the gdi track paths are the current platform ones for instance, or whatever the thing was with the spaces being (needlessly?) erased that another bug in RA was complaining about.

LoungeKatt:
Take a minute to actually read the replies and you may notice that changing the variable type was questioned, not the value. I would have a lot more time for other issues if you would just answer that question and stop trying to explain the entire purpose of the PR.

i30817:
I'm fine with you changing back the variable because there was a problem on visual studio. I'm not fine with neutering the bug fixes and reintroducing the buffer overflow on linux that's all. 

I'm explaining the purpose because on the commit you made to resolve the variable issue you found you actually reverted all the fixes.

LoungeKatt:
`#ifdef WIN32` means on Windows. It was explained by "There is no disagreement that on a system without drive designation, that the lost path components are an issue. I also cannot confirm without testing that it won't also cause issues on windows with a drive designation (hence simply excluding windows to move the fix along)."

i30817:
You know what, i'm just going to open another PR with the alternative code since this is getting messy. If you have commit rights you then can merge whichever one you feel best.

LoungeKatt:
I already moved to https://github.com/reicast/reicast-emulator/pull/1098 to fix it because I couldn't rewrite the existing PR remotely without making a bigger mess.

LoungeKatt:
If you have a different version, you are willing to submit that instead. The version at that URL above simply defines ssize_t for Windows. With any luck, it won't cause issues with `C:\game.gdi` and everyone will be happy.

i30817:
That's fine too although it's different from what i proposed to RA now... I'll close it real quick i guess.

PR #1098, opened 2018-04-02
fix gdi track parsing in unix #1092

LoungeKatt:
Submitted as a new pull, as rewriting history on a PR is outside the scope of the web interface.


skmp:
I really dislike the use of platform specific #defines and such, and the original use for C strings is quite a bit bug prone. Can the offending code be re-written to use C++ strings instead + find last of?

skmp:
@i30817 I never considered relative paths when I wrote that code %)

i30817:
you probably can replace it with lastof but you'd probably want the use something like `max(path.lastof('\\'), path.lastof('/')) + 1`

I don't actually know C++ but i noticed  that there was already some c++ isms in the file.

skmp:
@i30817 @LoungeKatt I don't have any gdis to test with right now, would be great if you can test it

skmp:
@i30817 got a moment to test this? Would love to merge

i30817:
Sorry i was busy with a few PRs (and i don't actually build reicast, the commit was for the RA fork). 

But i don't think i'll be a help. For one thing standalone RA just crashes from the command line if i build it myself with another error, so i can't actually test.

That said, the function looks ok, but it will lose the 'quality' of being able to parse out a relative path from windows in linux or vice versa, but that was only relevant for tracks in gdi with directories that were created on windows and moved to linux or vice versa. And since tosec doesn't believe in subdirectories it's unlikely to happen in practice.

skmp:
I'm not really familiar with retroarch so can't comment on that. Can you provide a couple of sample .gdi files (without the track files, of course) to test this?

i30817:
No it's a different failure from this code, it goes much further than loading a gdi without a track (which i did just now).

i30817:
also subtrack path handling with forward slash on linux works but as expected \ doesn't.
GDI : 3 tracks
file[1] "a/cs01.bin": FAD:0, CTRL:4, SSIZE:2352, OFFSET:0
file[2] "a/cs02.raw": FAD:450, CTRL:0, SSIZE:2352, OFFSET:0
file[3] "a/cs03.bin": FAD:45000, CTRL:4, SSIZE:2352, OFFSET:0
OSD: 6

LoungeKatt:
@skmp Are you leaving the test files in or was that just for verification?

skmp:
@LoungeKatt I'd like to start having tests for the code so I'll leave the test files in. They contain no data so it is not a copyright issue. 

@i30817 I've added path normalization for the gdi path, both on windows and linux, so that either `/` or `\` work as path separators.

@i30817 @zradoslav would be great if you can add some more 'exotic' .gdi files (with 0-byte track files of course) for future test cases.

LTGM?

PR #1099, opened 2018-04-02
Core changes for Android Studio conversion

LoungeKatt:


LoungeKatt:
travis is going to eventually need to update or will cause issues. The new Android Studio gradle plugin complains that no buildToolsVersion should be specified because each gradle version has a corresponding build tools, but travis requires specifying the build tools in their config. This means we either remove it locally to avoid the warning and risk travis failing for the wrong version on an update or leave the parameter and deal with Android Studio complaining all the time.

skmp:
(I know I've been taking my time reviewing this, past weekends have been too much party and hangover, an not enough reicast :<)

LoungeKatt:
travis-ci/pr:
`The command "GIT_BUILD=`git describe --all`-$GIT_HASH" failed and exited with 128 during .`

wercker:
`../../core/emitter/x86_emitter.h:241:5: error: 'template<class Rv, class ... Args> x86_ptr_imm::x86_ptr_imm(Rv (*)(Args ...))' cannot be overloaded
     x86_ptr_imm(Rv(DYNACALL * ptr)(Args...))`
`../../core/emitter/x86_emitter.h:234:2: error: with 'template<class Rv, class ... Args> x86_ptr_imm::x86_ptr_imm(Rv (*)(Args ...))'
  x86_ptr_imm(Rv(* ptr)(Args...))`

LoungeKatt:
Closing in favor of upcoming rebase on master.

PR #1104, opened 2018-04-11
tactx: Actually use aligned alloc functions

skmp:
Looks I somehow reverted half of the useful changes on the previous merge, duh. I need to find some nice visual diff tool for linux >.>

@hissingshark this should be alright now


LoungeKatt:
https://github.com/shiftkey/desktop if you want something "official"

PR #1107, opened 2018-04-11
CI Testing: From another repo

skmp:


LoungeKatt:
@skmp This brings about an interesting question. What do we do when there are legacy changes that would be grandfathered into the CLA, but the person is unavailable / unable to sign them?

I am thinking about certain changes from a certain third-party build that date back to 2014 / 2015

PR #1108, opened 2018-04-11
Yet another CI test: Noop change

skmp:


LoungeKatt:
The documentation for this sort of thing is very inconsistent with actual usage. I had the same issue trying to use the `[[]] ||` for keys. I ended up just not trying to use my own lol.

skmp:
Using `test /invesre cond/ || action` seems to work well enough for now (?)

PR #1133, opened 2018-04-21
Update libpng to 1.6.34

AbandonedCart:


skmp:
@LoungeKatt all of the build failtures :<. Shal we close this and take a second stab after #1206?

PR #1135, opened 2018-04-21
Snapcraft: Add .desktop path

skmp:


AbandonedCart (approved):
Marvelous

skmp:
Lol. The home folder is mounted elsewere under snapd, so we need to make the bios setup more user friendly. Otherwise, it seems to work....

PR #1161, opened 2018-06-02
Updated EmulatorView.mm

brand175:
adds dpad (joystick alt) controls. This won't work is the PadVeiwController.xib file isn't added to the project.


skmp:
@brand175 any idea why the build is failing? Also, can you add both files in a single MR? Thanks :)

LoungeKatt:
@skmp This is from https://github.com/reicast/reicast-emulator/issues/1145 and was replacing the joystick with a dpad simply to avoid resolving the joystick.

brand175:
@skmp I think it would be better not to merge the PR. Since it would be better if the joystick wasn’t a d-pad. I’ll redo the Touchscreen Triggers PR though.

skmp:
Closing in favor of #1221.

PR #1165, opened 2018-06-07
Fix video choppiness

flyinghead:
The YUV texture address was being initialized with the previous value instead of the current one.


skmp:
alololollololoolololokolol i’m so stupid xD xD

skmp:
Great catch. Wonder how that one happened. And I soend so many hours suspecting audio timing xD

flyinghead:
I did spend a day at least searching in the same direction, and gave up.
Then I noticed a random texture corruption in a game this morning, caused by YUV data being written at the wrong place in vram. That was it!

AbandonedCart (approved):

PR #1180, opened 2018-06-13
Removed unused code in core/linux-dist/main.cpp

baka0815:
This is a leftover from #964


LoungeKatt:
I'm trusting your judgement, since it passes and you were active in that issue ;)

PR #1181, opened 2018-06-13
X11: Refactored and fixed buttons

baka0815:
Some keys were not mapped like left and right shoulder buttons which I mapped to "F" and "V".
I also added constants for the X11 keys (f. e. "KEY_ESC") and used those instead of magic number ("9").
Also to avoid confusion I added "X11: " before some messages to know if this came from the X11 or another backend.

### DIFF
--- a/core/linux-dist/x11.cpp
+++ b/core/linux-dist/x11.cpp
@@ -77,12 +77,12 @@ void input_x11_handle()
 			{
 				case KeyPress:
 				case KeyRelease:
-					if (e.type == KeyRelease && e.xkey.keycode == 9) // ESC button
+					if (e.type == KeyRelease && e.xkey.keycode == KEY_ESC)
 					{
-						die("death by escape key");
+						die("X11: death by escape key");
 					}
 #if FEAT_HAS_NIXPROF
-					else if (e.type == KeyRelease && e.xkey.keycode == 76) // F10 button
+					else if (e.type == KeyRelease && e.xkey.keycode == KEY_F10)
 					{
 						if (sample_Switch(3000)) {
 							printf("Starting profiling\n");
@@ -91,7 +91,7 @@ void input_x11_handle()
 						}
 					}
 #endif
-					else if (e.type == KeyRelease && e.xkey.keycode == 95) // F11 button
+					else if (e.type == KeyRelease && e.xkey.keycode == KEY_F11)
 					{
 						x11_fullscreen = !x11_fullscreen;
 						x11_window_set_fullscreen(x11_fullscreen);
@@ -99,6 +99,24 @@ void input_x11_handle()
 					else
 					{
 						int dc_key = x11_keymap[e.xkey.keycode];
+
+						if (e.xkey.keycode == KEY_F)
+						{
+							// Left shoulder button pressed (lt)
+							if (e.type == KeyPress)
+								lt[0] = 255;
+							else
+								lt[0] = 0;
+						}
+						else if (e.xkey.keycode == KEY_V)
+						{
+							// Right shoulder button pressed (rt)
+							if (e.type == KeyPress)
+								rt[0] = 255;
+							else
+								rt[0] = 0;
+						}
+
 						if (e.type == KeyPress)
 						{
 							kcode[0] &= ~dc_key;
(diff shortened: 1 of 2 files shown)


baka0815:
@skmp anything blocking this?
It is needed to play f. e. F355 using the keyboard, because otherwise there is no possibility to accelerate.

baka0815:
@skmp what's the reason this doesn't get merged?
Is there anything I should change?
If this won't get merged, I could just close this.

skmp:
Gah, I had a mid-way review and I forgot to press the right button. Duh. Let me add the comments again -.-

core/linux-dist/x11.cpp
```diff
@@ -91,14 +91,32 @@ void input_x11_handle()
...
-					else if (e.type == KeyRelease && e.xkey.keycode == 95) // F11 button
+					else if (e.type == KeyRelease && e.xkey.keycode == KEY_F11)
 					{
 						x11_fullscreen = !x11_fullscreen;
 						x11_window_set_fullscreen(x11_fullscreen);
 					}
 					else
 					{
 						int dc_key = x11_keymap[e.xkey.keycode];
+
```
skmp: I think these should go though the `x11_keymap` lookup. Dreamcast itself uses just 16 bits, so we can use special values for the triggers.

core/linux-dist/x11.cpp
```diff
@@ -107,8 +125,9 @@ void input_x11_handle()
 						{
 							kcode[0] |= dc_key;
 						}
+
+						printf("KEY: %d -> %d: %d\n", e.xkey.keycode, dc_key, x11_dc_buttons );
```
skmp: I'd comment out this one, or make it use some debug log and not printf to avoid console spam

core/linux-dist/x11.cpp
```diff
@@ -117,24 +136,35 @@ void input_x11_handle()
...
+	//   A
+	x11_keymap[KEY_S] = DC_BTN_X;
+	x11_keymap[KEY_X] = DC_BTN_A;
+	x11_keymap[KEY_D] = DC_BTN_Y;
+	x11_keymap[KEY_C] = DC_BTN_B;
 
-	x11_keymap[53] = DC_BTN_X;
-	x11_keymap[54] = DC_BTN_B;
-	x11_keymap[55] = DC_BTN_A;
+	// ???
```
skmp: These were missing buttons on the original controller. They were used by some "arcade" controllers that were released by ascii and such iirc.

core/linux-dist/x11.cpp
```diff
@@ -117,24 +136,35 @@ void input_x11_handle()
...
+	x11_keymap[KEY_E] = DC_BTN_D;
+
+	// Start button (triangle)
+	x11_keymap[KEY_RETURN] = DC_BTN_START;
 
 	/*
 	//TODO: Fix sliders
-	x11_keymap[38] = DPad_Down;
-	x11_keymap[39] = DPad_Down;
+	x11_keymap[KEY_A] = DPad_Down;
```
skmp: I'd introduce some special DC_BTN_* values here and use them. Maybe DC_VIRT_LT to make it clear they do not exist in the h/w?

skmp:
@baka0815 sorry about the delay, life has been busy %)

core/linux-dist/x11.cpp
baka0815: Should I use ``#if defined(_DEBUG)``?

core/linux-dist/x11.cpp
baka0815: I'm sorry, but I don't understand what you mean here.

baka0815: Ah, I think I get it.
Fixing the sliders below (``TODO: Fix sliders``) and use 
````c
if (dc_key == DC_BTN_DPAD_DOWN)
````
instead of
````c
if (e.xkey.keycode == KEY_F)
````

core/linux-dist/x11.cpp
baka0815: Are the sliders meant to be the shoulder buttons LT and RT?
Then why not just use ``DC_AXIS_LT`` and ``DC_AXIS_LT`` which are defined in https://github.com/reicast/reicast-emulator/blob/a4cf716f99ad8f76b85c24063b12958bdd82f22d/core/linux-dist/main.h#L31

core/linux-dist/x11.cpp
skmp: Yup!

baka0815: Ok, then this should be done.

skmp:
Yeah, I accidentally broke travis during the CD changes for external PRs. I'll add you as a collaborator as a workaround for now.

PR #1182, opened 2018-06-13
evdev: Allow for custom <device>.cfg files in the /mappings/ directory.

baka0815:
Check if a configuration file named exactly like the device exists in the /mappings/ directory.
If it does, use this instead of the generic one.

Exposes ``file_exists()`` from stdclass.

### DIFF
--- a/core/linux-dist/evdev.cpp
+++ b/core/linux-dist/evdev.cpp
@@ -223,7 +223,9 @@
 
 				if(custom_mapping_fname != NULL)
 				{
+					// custom mapping defined in config, use that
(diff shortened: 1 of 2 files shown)


skmp:
Looks good to me. @Holzhaus got any objections or follow-ups you want us to do?

danboid:
I'd like to see the README updated to cover joysticks. libevdev isn't mentioned as a build requirement for Linux although the majority of users will want joystick support. Also, the way it works (ie what name to use for the config file etc) should be documented too.

skmp:
@danboid want to open a pull request with the proposed changes?

PR #1183, opened 2018-06-14
README: Revert to irc from slack

skmp:
We should set up a slack <-> irc bridge, too.


skmp:
I personally use IRCCloud which is similar to slack anyway

skmp:
Any objections?

LoungeKatt:
I honestly don't have any objections, since I use whatever when necessary.

AbandonedCart (approved):

PR #1185, opened 2018-06-17
[c++11] core: adding high resolution clock

zradoslav:
Changing _os_GetSeconds_ code to platform agnostic, enabling high resolution clock.
Also removed redundant includes.

### DIFF
--- a/core/hw/sh4/dyna/driver.cpp
+++ b/core/hw/sh4/dyna/driver.cpp
@@ -18,7 +18,6 @@
 #include "hw/aica/aica_if.h"
 #include "hw/gdrom/gdrom_if.h"
 
-#include <time.h>
 #include <float.h>
 
 #include "blockmanager.h"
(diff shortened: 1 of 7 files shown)


LoungeKatt:
It seems chrono is not supported by the Android NDK

skmp:
I would not trust many environents having a good clock implementation. Maybe behind an #ifdef for unix and we wait for it to mature?

zradoslav:
@skmp 
So you recommend just to rewrite UNIX (not ANDROID) part in such way, leaving Windows unchanged?

skmp:
I would make a global define - USE_CPP_TIME - and use #ifs to toggle the code. Or even HOST_TIME and TIME_CPP / TIME_NATIVE (same pattern as the other config options)

skmp:
(Cleaning up old tickets)

PR #1211, opened 2018-07-09
(Naomi) Fix for "The Rumble Fish 2"

barbudreadmon:
I think this value is the SRAM size, naomi had 64KB SRAM (source: https://segaretro.org/Sega_NAOMI)

### DIFF
--- a/core/types.h
+++ b/core/types.h
@@ -238,7 +238,7 @@ struct vram_block
 	#define VRAM_SIZE (16*1024*1024)
 	#define ARAM_SIZE (8*1024*1024)
 	#define BIOS_SIZE (2*1024*1024)
-	#define BBSRAM_SIZE (8*1024)
+	#define BBSRAM_SIZE (64*1024)
 
 	#define ROM_PREFIX "naomi_"
 	#define ROM_NAMES ";epr-21576d.bin"


skmp:
According to make, naomi is 8k, aw is 64. Is this a aw rom?

barbudreadmon:
It is an aw rom converted to naomi.

barbudreadmon:
@skmp So, what are you going to do about the support for this "naomiswave" game ? It seems increasing sram size is necessary for it. Are you sure about the sram size being 8k in mame (i didn't find this information) ? I checked on google and found multiple reference to naomi having 64KB as sram size (most seemed to be copy/paste of the link i posted above though).

barbudreadmon:
Actually, i have seen @p1pkin (the author of demul, which is also MetalliC from mame i think) saying NAOMI/NAOMI2 have "32KB for game NVdata directly accessible by CPU", is it the same as SRAM ?

p1pkin:
@barbudreadmon correct, N1/2 have 32KByte battery powered SRAM.
Atomiswave have it 128Kbyte, so I'd imagine some of AW->N conversions attempting to use more SRAM than N really have.

barbudreadmon:
@p1pkin Thanks for the confirmation ! 32KB seems to be enough for this game actually, so i'll create a new PR with the correct value.

Issue #1150, opened 2018-05-17
Reicast @ Hack4Glarus summer edition (06~08.07.2018)


skmp:
I will be participating at the [Hack4Glarus summer edition hackathon](https://www.facebook.com/events/199003247383696/) with reicast as a project. I know some other reicast members might also join, and wanted to make an open call.

If you are interested in joining, please fill in the [form]( https://docs.google.com/forms/d/e/1FAIpQLSeku90cUufMGJqdvkXUWaGNUqqcymsHT5i1A8y03Tv1y7AeOw/viewform) and put "reicast" as the project, and "reicast users" as to who will be benefiting. You can of course apply with your own projects, or work on multiple things during the weekend. Thanks :)

Remote participation
---

You can join us remotely, either via [#hack-4-glarus in ungleich mattermorst](https://chat.ungleich.ch/ungleich/channels/hack-4-glarus) or [#reicast in freenode (IRCCloud link)](https://www.irccloud.com/invite?channel=%23reicast&hostname=irc.freenode.net&port=6667)

Hackathon ideas
---

Done
- [x] Minimal cmodel implementation that runs 256b
- [x] Investigate j2/j4 cores
- [x] Implement SMC for x64 / rec-cpp & related easy fixes (#1168, #1167, #1141, #380, #107)
- [x] Investigate https://cmty.app/ (Applied, Waiting for reply)
- [x] Install vivado and web ISE
- [x] Get the zynq dev board running a simple design

To do
- [ ] progress report on site, donations update for site
- [ ] Compile aarch64 snaps via lincpp
- [ ] Per game ini loading
- [ ] Make the debug checks optional for x86
- [ ] Look for other bounty systems apart from bountysource
- [ ] Work in code of conduct / donations / purpose documentation and legalese stuff
- [ ] Make a bounty dashboard page
- [ ] Logos, banners and visual design for fb and such
- [ ] consolidate information for new devs / attract devs

r7 release push
- [x] #1109  in armv7 core
- [ ] gamedb organizational work
- [x] "fast track" continuous delivery of post r7 beta ("Reicast Edge" or something?)
- [ ] Add donation information in android shell

Maybe for another day
- [ ] Minimal verilog implementation that runs 256b
- [ ] aarch64 backend
- [ ] Add http streaming for disks in reicast.js
- [ ] libdreamcast brainstorming
- [ ] Forums url link and such (coordinate w/ |darc| about it?)
- [ ] Integrate test suites from redream and lxdream
[...]

PR #1212, opened 2018-07-09
README: Bug reporting and Discord link

LoungeKatt:


skmp:
We need to also add gamedb and more stuff, I'm linking #1150 so that I will follow up on that one.

PR #1213, opened 2018-07-09
Issues: Fix formatting and remove default image

AbandonedCart:


LoungeKatt:
No reason to have every issue include some Github icon

baka0815:
@AbandonedCart we should port this to the gamedb project.

AbandonedCart:
Port what? This was pulling box art from a website to act as game icons. Supposedly, images for the games can be pulled directly from the ROM. That would still beat any other website source.

AbandonedCart:
Right. Was thinking of the recent https://github.com/reicast/reicast-emulator/commit/b56d0f09c933c76fcf4058e34e9888cec3a134a3 not the commit from over a month ago.

AbandonedCart:
If you want to port it, by all means. If you want me to port it, then it will have to get in line behind a long list of other things already in the works.

PR #1197, opened 2018-07-03
LGTM: Fix "Function input_evdev_handle should return a value of type bool but does not return a value here"

baka0815:
The return value of input_evdev_handle() was never used, so just return void.

This fixes one of the lgtm warnings (see #1117).

PR #1222, opened 2018-07-15
Fix controller support for Linux when building with GCC8

t0pu:
On Linux the window stays black without any error message when a joystick/controller is connected and reicast was build with GCC8. This is probably due to changes made in GCC8. https://gcc.gnu.org/gcc-8/porting_to.html#Wreturn-type

### DIFF
--- a/core/linux-dist/evdev.cpp
+++ b/core/linux-dist/evdev.cpp
@@ -468,6 +468,7 @@
 					break;
 			}
 		}
+		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


baka0815:
The return value of ``input_evdev_handle`` is never used (see #1197 which changes this method to ``void``).

t0pu:
I didn't see your PR @baka0815 but changing `input_evdev_handle` to `void` would be a even better solution if the return value is never used. So merging #1197 would also fix the controller problem with GCC8.

PR #1232, opened 2018-07-23
7oxicshadow linux clean exit

flyinghead:
Updated PR #917 so it can be auto merged. These changes are badly needed.

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -26,19 +26,19 @@ ifndef NO_REC
 endif
 
 ifndef NOT_ARM
-    RZDCY_MODULES += rec-ARM/
+    RZDCY_MODULES += rec-ARM/ rec-common/
 endif
 
 ifdef X86_REC
-    RZDCY_MODULES += rec-x86/ emitter/
+    RZDCY_MODULES += rec-x86/ emitter/ rec-common/
 endif
 
 ifdef X64_REC
-    RZDCY_MODULES += rec-x64/
+    RZDCY_MODULES += rec-x64/ rec-common/
 endif
 
 ifdef CPP_REC
-    RZDCY_MODULES += rec-cpp/
+    RZDCY_MODULES += rec-cpp/ rec-common/
 endif
 
 ifndef NO_REND
--- a/core/hw/pvr/Renderer_if.cpp
+++ b/core/hw/pvr/Renderer_if.cpp
@@ -229,6 +229,8 @@ bool rend_single_frame()
 	return do_swp;
 }
 
+int rend_en = true;
+
 void* rend_thread(void* p)
 {
 #if FEAT_HAS_NIXPROF
@@ -269,11 +271,19 @@ void* rend_thread(void* p)
 	//we don't know if this is true, so let's not speculate here
 	//renderer->Resize(640, 480);
 
-	for(;;)
+	while(rend_en)
 	{
 		if (rend_single_frame())
 			renderer->Present();
 	}
+	
+	return (0);
+}
+
+void rend_terminate()
+{
+	rend_en = false;
+	printf("rend_terminate called\n");
 }
 
 #if !defined(TARGET_NO_THREADS)
--- a/core/linux-dist/evdev.cpp
+++ b/core/linux-dist/evdev.cpp
@@ -321,6 +321,8 @@
 		}
 	}
 
+        void start_shutdown(void);
+
 	bool input_evdev_handle(EvdevController* controller, u32 port)
 	{
 		#define SET_FLAG(field, mask, expr) field =((expr) ? (field & ~mask) : (field | mask))
@@ -353,7 +355,7 @@
 					} else if (ie.code == controller->mapping->Btn_Start) {
 						SET_FLAG(kcode[port], DC_BTN_START, ie.value);
 					} else if (ie.code == controller->mapping->Btn_Escape) {
-						die("death by escape key");
+                                                start_shutdown();
 					} else if (ie.code == controller->mapping->Btn_DPad_Left) {
 						SET_FLAG(kcode[port], DC_DPAD_LEFT, ie.value);
 					} else if (ie.code == controller->mapping->Btn_DPad_Right) {
--- a/core/linux-dist/main.cpp
+++ b/core/linux-dist/main.cpp
@@ -222,6 +222,7 @@ void os_DoEvents()
 {
 	#if defined(SUPPORT_X11)
 		input_x11_handle();
+		event_x11_handle();
 	#endif
 }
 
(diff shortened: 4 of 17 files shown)


core/core.mk
```diff
@@ -26,19 +26,19 @@ ifndef NO_REC
 endif
 
 ifndef NOT_ARM
-    RZDCY_MODULES += rec-ARM/
+    RZDCY_MODULES += rec-ARM/ rec-common/
```
baka0815: Is there a case where we don't want to add ``rec-common``?

core/hw/pvr/Renderer_if.cpp
```diff
@@ -229,6 +229,8 @@ bool rend_single_frame()
 	return do_swp;
 }
 
+int rend_en = true;
```
baka0815: I would like to have a comment here or a more meaningful name (what does the ``_en`` mean?).
Maybe we should call it ``rend_end`` or ``rend_done`` and use ``while (!rend_end/done)`` below?

core/linux-dist/evdev.cpp
```diff
@@ -321,6 +321,8 @@
 		}
 	}
 
+        void start_shutdown(void);
```
baka0815: Indentation.

core/linux-dist/evdev.cpp
```diff
@@ -353,7 +355,7 @@
 					} else if (ie.code == controller->mapping->Btn_Start) {
 						SET_FLAG(kcode[port], DC_BTN_START, ie.value);
 					} else if (ie.code == controller->mapping->Btn_Escape) {
-						die("death by escape key");
+                                                start_shutdown();
```
baka0815: Indentation.

I would also like to add the string to ``start_shutdown()`` to differntiate between ``"death by escape key"`` (evdev) and ``"X11: death by escape key"`` (X11). But that's a minor nit and I'm fine either way.

core/linux-dist/main.cpp
```diff
@@ -475,11 +501,38 @@ int main(int argc, wchar* argv[])
...
+	
 	#ifdef TARGET_PANDORA
 		clean_exit(0);
 	#endif
 
+        #if HOST_OS==OS_LINUX
+
+            printf("main loop ended\n");
+
+            dc_term();
```
baka0815: Too many newlines.

core/linux-dist/main.cpp
```diff
@@ -475,11 +501,38 @@ int main(int argc, wchar* argv[])
...
+
+	#if defined(USE_EVDEV)
+            printf("closing any open controllers\n");
+
+            for (int port = 0; port < 4 ; port++)
+            {
+                if(evdev_controllers[port].fd >= 0)
```
baka0815: Spacing: ``if (``

core/linux-dist/main.cpp
```diff
@@ -475,11 +501,38 @@ int main(int argc, wchar* argv[])
...
+	#endif
+
+	#if defined(SUPPORT_X11)
+		/*Close the GL context */
+		x11_gl_context_destroy();
+		/* Destroy the window */
+		x11_window_destroy();
+	#endif
+
+        #endif
```
baka0815: Indentation.

core/linux-dist/x11.cpp
```diff
@@ -64,6 +66,23 @@ void x11_window_set_fullscreen(bool fullscreen)
...
+
+void event_x11_handle()
+{
+	XEvent event;
+	
+	while(XPending(x11_disp)) {
+		XNextEvent(x11_disp, &event);
+
+		if (event.type == ClientMessage &&
+		event.xclient.data.l[0] == wmDeleteMessage) {
```
baka0815: Indentation?

core/linux-dist/x11.cpp
```diff
@@ -64,6 +66,23 @@ void x11_window_set_fullscreen(bool fullscreen)
 		XSendEvent((Display*)x11_disp, DefaultRootWindow((Display*)x11_disp), False, SubstructureNotifyMask, &xev);
 }
 
+void start_shutdown(void);
+
+void event_x11_handle()
+{
+	XEvent event;
+	
+	while(XPending(x11_disp)) {
```
baka0815: Spacing: ``while (``

core/nullDC.cpp
```diff
@@ -143,15 +143,24 @@ int dc_init(int argc,wchar* argv[])
...
+
+	// Command line parsing moved after cfgopen as it allows
+	// the command line to make changes to the "settings" after
+	// they have been read. This helps with the NO_VIRTUAL_CFG
+	// option.
+	// After the command line has been parsed the settings are
+	// loaded using LoadSettings().
+
+	if(ParseCommandLine(argc,argv))
+        {
```
baka0815: Indentation.

core/hw/pvr/Renderer_if.cpp
skmp: this stands for rendering enabled, iirc it's the control flag of the rendering thread?

skmp:
Why is `NO_VIRTUAL_CFG` support a requirement? My understanding is that it was some workaround for a bug that we fixed a few months back?

core/windows/winmain.cpp
```diff
@@ -327,9 +332,11 @@ LRESULT CALLBACK WndProc2(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
 		}
 		// Handles the close message when a user clicks the quit icon of the window
 	case WM_CLOSE:
+	{
+		start_shutdown();
```
skmp: This should be moved to `WM_QUIT` iirc. `WM_CLOSE` should just `PostQuitMessage()` for windows.

core/rec-x86/rec_x86_asm.cpp
```diff
@@ -89,7 +91,11 @@ naked void ngen_mainloop(void* cntx)
 		jnz do_iter;
 		ret;
 
-do_iter:
+	do_iter:
```
skmp: It would be faster here to use the `return-to-fragment-via-rv` approach to avoid paying the cost of the if in the main loop. I guess that's good enough for now, and on x86 we can eat the perf hit anyway

core/rec-x86/rec_lin86_asm.S
```diff
@@ -134,6 +141,7 @@ cleanup:
 	pop esi
 	pop esi
 	pop esi
+	pop esi             # This additional pop is required otherwise a segfault occurs??
```
skmp: This is fishy

core/rec-common/rec_common_helper.cpp
```diff
@@ -0,0 +1,9 @@
+#include "types.h"
+
+unsigned int ngen_required = true;
```
skmp: How would this work with the arm JITs? We can't reference global symbols there because of ... linking issues.

PR #1233, opened 2018-07-23
For alpha testing punch-through polygon, use the alpha value from the…

flyinghead:
… texture, not vertex.

Fixes Jet Grind Radio glitches with PT polygons (bus, trucks)
Seems to happen only on NVidia


flyinghead:
Here is what the bug looks like:
jet grind radio pt alpha issue

flyinghead:
Then it could be a rounding error that causes the alpha to fluctuate around the alpha test value. I'll have a look.

Issue #813, opened 2015-08-27
Add Windows release builds *.exe


ofry:
Hello.

Thanks for adding Windows builds artifacts, but I can't run *.exe without installed Visual Studio, because this exe required MSVCP120D.dll

Please add release build :)

Thanks.

skmp:
We cannot do release builds because they take too much time. I'll convert
the projects to static link though

On Thu, Aug 27, 2015 at 11:23 AM, ofry <notifications@github.com> wrote:

> Hello.
>
> Thanks for adding Windows builds artifacts, but I can't run *.exe without
> installed Visual Studio, because this exe required MSVCP120D.dll
>
> Please add release build :)
>
> Thanks.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/reicast/reicast-emulator/issues/813>.
>



-- 
~skmp

skmp:
Okay, done, try out 435f104 or wait for ed143ec to build

ofry:
ed143ec: same problem (required VCOMP120D.dll).

skmp:
Hmm, openmp can only be linked dynamically. Oh well

On Thu, Aug 27, 2015 at 11:49 AM, ofry <notifications@github.com> wrote:

> ed143ec
> <https://github.com/reicast/reicast-emulator/commit/ed143eccc9167f9158ae2ac5d7a5a66d0764793a>:
> same problem (required VCOMP120D.dll).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/reicast/reicast-emulator/issues/813#issuecomment-135367548>
> .
>



-- 
~skmp

skmp:
Does https://www.microsoft.com/en-us/download/details.aspx?id=40784 help?

ofry:
No, this doesn't help. I downloaded and installed x86 and x64 versions, rebooted PC, and tried to launch reicast. Same error.

OS: Windows 8 x64

williansti:
I uploaded what you need, make your tests! http://www.mediafire.com/download/103l92cdovvud62/Reicast+2015-09-04+for+windows+x64.zip

ofry:
Thanks. It's workinig, but it's still debug build :)

ofry:
Maybe try to connect with Orphis? His buildbot already compiling DeSmuMe, PPSSPP, PCSX2 and JPSCP.

ofry:
http://buildbot.orphis.net/

[...]

PR #1234, opened 2018-07-23
GL3: Use log2 depth scaling in fragment shader

flyinghead:
Currently a compile time option (!GLES). Fixes many depth issues (Dynamite Cop, Cannon Spike, etc.)

### DIFF
--- a/core/rend/gles/gles.cpp
+++ b/core/rend/gles/gles.cpp
@@ -100,10 +100,21 @@ void main() \n\
 	vec4 vpos=in_pos; \n\
 	vtx_xyz.xy = vpos.xy;  \n\
 	vtx_xyz.z = vpos.z*sp_FOG_DENSITY;  \n\
-	vpos.w=1.0/vpos.z;  \n\
+	vpos.w=1.0/vpos.z;  \n"
+#ifndef GLES
+	"\
+	if (vpos.w < 0.0) { \n\
+		gl_Position = vec4(0.0, 0.0, 0.0, vpos.w); \n\
+		return; \n\
+	} \n\
+	vpos.z = vpos.w; \n"
+#else
+	"\
+	vpos.z=depth_scale.x+depth_scale.y*vpos.w;  \n"
+#endif
+	"\
 	vpos.xy=vpos.xy*scale.xy-scale.zw;  \n\
 	vpos.xy*=vpos.w;  \n\
-	vpos.z=depth_scale.x+depth_scale.y*vpos.w;  \n\
 	gl_Position = vpos; \n\
 }";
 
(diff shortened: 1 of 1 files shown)


core/rend/gles/gles.cpp
```diff
@@ -340,8 +356,13 @@ const char* ModifierVolumeShader =
 uniform lowp float sp_ShaderColor; \n\
 /* Vertex input*/ \n\
 void main() \n\
-{ \n\
-	" FRAGCOL "=vec4(0.0, 0.0, 0.0, sp_ShaderColor); \n\
+{ \n"
+#ifndef GLES
+	"\
+	highp float w = gl_FragCoord.w * 100000.0; \n\
+	gl_FragDepth = log2(1.0 + w) / 34.0; \n"
```
skmp: I guess those numbers need some tuning from time to time? I used a (somewhat) similar technique for nullDC, though clamped the exponent part to make sure everything fits - https://github.com/skmp/nulldc/blob/master/nulldc/plugins/drkPvr/ps_hlsl.fx#L143

flyinghead: Yes, redream was using log2(1 + w * 100) / 24 originally.
I bumped it up as I found one case of loss of precision on very small z.

skmp:
Alright. We really should make an option out of this as it breaks MSAA on desktop, at some point.

skmp:
This is GL3.x isn't it? I don't recall if the GL device is 3.x or 4.x

skmp:
MSAA depends on per sample depth, the fragment shader runs.. per fragment. so in 16x samples end up with the same depth, effectivelly disabling it.

In msaa modes the depth buffer is 2-16x the density of the pixel buffer, to better resolve edges

blackman91:
Nice! So this finally fixes Cannon Spike on Android too?

Deivmsr:
I tested (055c359) the last beta of Google Play and Cannon Spike continued with the same problems. I do not know if any option should be activated.

blackman91:
Yeah just tested the same version and no combination of settings fixes it, the bug still present, I guess this only works on the pc version:
screenshot_20180723-160154

blackman91:
I'll try to test the game on standalone pc reicast .

blackman91:
The master builds are uninstallable on windows for some reason, has anyone been able to successfully install them?

blackman91:
Nevermind I figured how to install them https://github.com/reicast/reicast-emulator/issues/813

I tested Cannon Spike and it works on the windows build!
cannonspikeworking

So it only works on pc, any chance you guys could port these changes to gles3 so I could be playable on Android?

Deivmsr:
blackman91, In the last build of fh / mymaster of flyinghead were repaired the broken graphics of Cannon Spike and the shadow of Sonic Adventure 2, where even allowed me to pass the boss King Boom Boo boss. But this was only possible because the flyinghead added a system that automatically identifies the GLES version that the device uses.

blackman91:
Holy shit you are right, Cannon Spike is finally playable after all these years on mobile on flyinghead branch!!! Amazing job as usual, it even fixes a lot of bugs in other games like EGG and the performance is perfect even on motog5:
screenshot_20180807-212943

Hope these changes come to the main branch because sadly this branch still has the cdi blackscreen bug.

blackman91:
Wow that build even fixes the graphics errors of a lot of games like Seaman, Rayman 2, Jedi Power Battles, Zombie Revenge, EGG without performance hit:
https://github.com/reicast/gamedb/issues/7

https://github.com/reicast/gamedb/issues/8

https://github.com/reicast/gamedb/issues/9

https://github.com/reicast/gamedb/issues/10

https://github.com/reicast/gamedb/issues/11

Really impressive fh.

PR #1235, opened 2018-07-23
Float math in alpha calculation causes some punch-through textures to…

flyinghead:
… have alpha noise.

### DIFF
--- a/core/rend/gles/gles.cpp
+++ b/core/rend/gles/gles.cpp
@@ -336,7 +336,7 @@ void main() \n\
 	} \n\
 	#endif\n\
 	#if cp_AlphaTest == 1 \n\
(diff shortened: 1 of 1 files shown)


skmp:
I think the real bug is in https://github.com/reicast/reicast-emulator/blob/055c3599b6cfa99279542dc2d29c6e5032995ded/core/rend/gles/gles.cpp#L1713, should be `/256.0f`, not `/255.0f`

skmp:
I think in general, we should be using `256` as `1.0` not `255`

flyinghead:
Isn't GL going to saturate to 1.0 for a value of 255 in an 8-bit channel? If that's the case then we should do the same.
I think the issue here is that we were using float math with >, which gives different results than integer math.

AbandonedCart:
The added changes can be added later. As it stands now, there are some serious issues with alpha.

PR #1237, opened 2018-07-25
Android: Background service (Part 1)

AbandonedCart:
This is a counterpart to the future support for pausing the core.

It moves termination processes from stop to destroy and instantiates a service that will keep the application active in memory when not active on the device.


blackman91:
Amazing, nice job man it works great!

AbandonedCart:
Nothing actually changed yet.

blackman91:
What do you mean? I can now continue playing without the app restarting when I press the home button or open another app on latest beta.

AbandonedCart:
Interesting. The service isn't actually enabled. Could be dump luck that this combined with the "proper close" changes by @flyinghead let it run for a short time. It will still eventually get killed off by the system, though.

blackman91:
Lucky, I was playing MvsC2 then started browsing on Firefox for Android for like 20 minutes then came back to reicast and it still resumed correctly.

PR #1238, opened 2018-07-26
Fix to windows OpenFile dialog

pold500:
Now it actually says something about what exactly it tries to open, also on "cancel" it exits gracefully.

Proposal: instead of checking one thousand return codes, which are easily ignored (nobody checks return codes from the functions, as it seems), I suggest to introduce an exceptions (this part of emulator doesn't require execution speed, so exceptions here would be just fine).


core/imgread/common.cpp
```diff
@@ -188,16 +188,7 @@ bool InitDrive(u32 fileflags)
 #else
 	int gfrv=0;
```
skmp: I would also init here to some symbolic `rv_*` value just to make the code nicer

core/imgread/common.cpp
```diff
@@ -211,11 +202,11 @@ bool InitDrive(u32 fileflags)
 	if (!InitDrive_(fn))
```
skmp: Why not rewrite this as `if (InitDrive(fn)) return true; else { /* no disk handling code*/; return false; }`? I think it would make the code easier to read.

core/nullDC.cpp
```diff
@@ -39,35 +39,25 @@ settings_t settings;
 #include <windows.h>
 #endif
 
-int GetFile(char *szFileName, char *szParse=0,u32 flags=0) 
+int GetFile(char *szFileName, char *szParse = 0, u32 flags = 0)
 {
-	cfgLoadStr("config","image",szFileName,"null");
-	if (strcmp(szFileName,"null")==0)
+#if HOST_OS==OS_WINDOWS
+	cfgLoadStr("config", "image", szFileName, "");
```
skmp: cfgLoadStr should not be within the windows `#if` check as it applies to all platforms

core/imgread/common.cpp
baka0815: No need for an else if we return from the true case. This would result in even less nesting.

core/imgread/common.cpp
```diff
@@ -164,131 +171,98 @@ bool InitDrive_(wchar* fn)
...
-		printf("Loading default image \"%s\"\n",settings.imgread.DefaultImage);
-		if (!InitDrive_(settings.imgread.DefaultImage))
+		printf("Loading default image \"%s\"\n", settings.imgread.DefaultImage);
+		if (!InitDrive_(settings.imgread.DefaultImage.c_str()))
 		{
-			msgboxf("Default image \"%s\" failed to load",MBX_ICONERROR);
+			msgboxf("Default image \"%s\" failed to load", MBX_ICONERROR);
 			return false;
 		}
 		else
```
baka0815: Remove the ``else`` please, it just creates clutter as it's unnecessary here.

core/imgread/common.cpp
```diff
@@ -164,131 +171,98 @@ bool InitDrive_(wchar* fn)
...
+		SetSnsCodes(0x28, 0x00, 0x06);
 		return true;
 	}
-	else if (gfrv == -1)
+	else
 	{
-		sns_asc=0x28;
-		sns_ascq=0x00;
-		sns_key=0x6;
+		SetSnsCodes(0x28, 0x00, 0x06); // to fix: we have same return codes as previous branch, looks suspicious
```
baka0815: Leave a ``TODO`` here instead of "to fix" (``// TODO we have...``).

Maybe it should be ``SetSnsCodes(0x29, 0x00, 0x06);`` as in ``InitDrive()?`` (line 209).

core/imgread/common.cpp
```diff
@@ -164,131 +171,98 @@ bool InitDrive_(wchar* fn)
...
+		SetSnsCodes(0x28, 0x00, 0x06);
 		return true;
 	}
 	else
 	{
-		sns_asc=0x28;
-		sns_ascq=0x00;
-		sns_key=0x6;
-		return true;
+		SetSnsCodes(0x28, 0x00, 0x06);
```
baka0815: Same ``TODO`` as above as it's the same code in both branches?
Maybe it should be ``SetSnsCodes(0x29, 0x00, 0x06);`` as in ``InitDrive()?`` (line 209).

The return code was changed from ``true`` to ``false`` was this intentional? (I think it was just asking to be sure)

core/nullDC.cpp
```diff
@@ -39,35 +39,38 @@ settings_t settings;
...
-			//strcpy(szFileName,ofn.lpstrFile);
+			const DWORD errorCode = CommDlgExtendedError();
+			if (errorCode == FNERR_BUFFERTOOSMALL)
+			{
+				//The first two bytes of the lpstrFile buffer contain an integer value specifying the size required to receive the full name, in characters.
+				const size_t newBufferSize = ((size_t)ofn.lpstrFile[0] | (size_t)ofn.lpstrFile[1] << 8);
+				diskFileName.resize(newBufferSize);
+				//try again
+			}
+			else return (errorCode == ERROR_SUCCESS) ? rv_ok : rv_error;
```
baka0815: Break after the else.

core/windows/winmain.cpp
```diff
@@ -687,7 +687,10 @@ int CALLBACK WinMain(HINSTANCE hInstance,HINSTANCE hPrevInstance,LPSTR lpCmdLine
 		int dc_init(int argc,wchar* argv[]);
 		void dc_run();
 		void dc_term();
-		dc_init(argc,argv);
+		if(dc_init(argc,argv)!=rv_ok)
```
baka0815: Spacing: ``if (dc_init(argc,argv) != rv_ok)``

core/nullDC.cpp
pold500: return does the break

baka0815: Sorry, I meant a line break not a break statement.

LoungeKatt:
This changes function parameters, and moves previously universal code into a Windows-specific block. While it may make the open dialog look nicer on Windows, that should not come at the cost of breaking every other platform.

pold500:
@LoungeKatt this has nothing to do with windows, std::string is quite cross-platform, looks like I just haven't updated all the necessary calls to the updated signature. I'll fix it.

core/nullDC.cpp
```diff
@@ -39,35 +39,38 @@ settings_t settings;
 #include <windows.h>
 #endif
 
-int GetFile(char *szFileName, char *szParse=0,u32 flags=0) 
+int GetFile(std::string& diskFileName)
 {
-	cfgLoadStr("config","image",szFileName,"null");
-	if (strcmp(szFileName,"null")==0)
```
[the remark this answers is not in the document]
baka0815: @LoungeKatt the null check is now inside thr #if, but there was no handling other than the Windows one previously, so no functionality removed here.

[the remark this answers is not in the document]
LoungeKatt: @baka0815 Correct, but https://github.com/reicast/reicast-emulator/blob/055c3599b6cfa99279542dc2d29c6e5032995ded/core/hw/naomi/naomi_cart.cpp#L209 should also have been changed to match.

core/windows/winmain.cpp
LoungeKatt: @baka0815 wouldn't that also mean needing to change `int dc_init(int argc,wchar* argv[]);` to `int dc_init(int argc, wchar* argv[]);` which is outside the scope of the actual changes?

skmp:
This pull request **fixes 4 alerts** when merging 1e5ca43af7319ae0c321979f3995ad7cb4521317 into 055c3599b6cfa99279542dc2d29c6e5032995ded - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-36e5d8a63ede4fe77093dc8fc4848112a7e1690e)

**fixed alerts:**

* 4 for FIXME comment

---

core/windows/winmain.cpp
baka0815: @LoungeKatt I would also say that that change is out of scope for this PR, but since @pold500 changed that one line, the spacing should be taken care of.

LoungeKatt:
This now fixes what it originally broke and the current Travis error (for debug key permissions) can be disregarded.

skmp:
I'll review further tonight, loads of changes I see.

skmp:
This pull request **fixes 4 alerts** when merging 82c0ef59f81c60fb81b898057cce6623d9827eb0 into 8d4f8597ef5fb8319db355f2d856ba51893908ec - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-27f6b15fd2a5b8d4633c7e0401e8e34788080106)

**fixed alerts:**

* 4 for FIXME comment

---

skmp:
This pull request **fixes 4 alerts** when merging c68c2935440ad2cf38de0df68112f95e9d24d292 into 51a0c5727f2992004418d079631101da944eb917 - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-2f5ac018251db7f8475d262154c9b8053eece056)

**fixed alerts:**

* 4 for FIXME comment

---

core/imgread/common.cpp
pold500: Fixed

skmp:
This pull request **fixes 5 alerts** when merging 7366ef03ef2dda1dba0cf2f73516e1dd3c21f03e into 51a0c5727f2992004418d079631101da944eb917 - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-c4857732ab744f53fd03ac7d4212d778a066f5ac)

**fixed alerts:**

* 5 for FIXME comment

---

core/hw/naomi/naomi_cart.cpp
```diff
@@ -204,13 +204,14 @@ bool naomi_cart_LoadRom(char* file)
...
+
 #if HOST_OS == OS_WINDOWS
-	if (strcmp(SelectedFile, "null") == 0) {
+	if (SelectedFile == "null")
+    {
 		OPENFILENAME ofn = { 0 };
 		ofn.lStructSize = sizeof(OPENFILENAME);
 		ofn.hInstance = (HINSTANCE)GetModuleHandle(0);
-		ofn.lpstrFile = SelectedFile;
+		ofn.lpstrFile = &SelectedFile[0];
```
skmp: why not `SelectedFile.c_str()`?

core/imgread/common.cpp
```diff
@@ -23,6 +24,19 @@ Disc*(*drivers[])(const wchar* path)=
 
 u8 q_subchannel[96];
 
+typedef std::tuple<int, int, int> SenseErrorCode;
```
skmp: why not simply use `typdef int SenseErrorCode[3]`?

skmp: Or even a struct w/ asc, ascq and key?

core/imgread/common.cpp
skmp: ? looks fine to me? Why do you think it should be removed?

core/imgread/common.cpp
```diff
@@ -138,158 +152,122 @@ Disc* OpenDisc(const wchar* fn)
...
 		else
 			return true;
 	}
 
-	// FIXME: Data loss if buffer is too small
-	wchar fn[512];
-	strncpy(fn,settings.imgread.LastImage, sizeof(fn));
-	fn[sizeof(fn) - 1] = '\0';
-
+	std::string diskImageFileName{ settings.imgread.LastImage };
```
skmp: Should be able to use `=` instead of `{ ... }` for initialization

core/imgread/common.cpp
```diff
@@ -138,158 +152,122 @@ Disc* OpenDisc(const wchar* fn)
...
-		gd_setdisc();
-		sns_asc=0x29;
-		sns_ascq=0x00;
-		sns_key=0x6;
-		return true;
-	}
-	else if (gfrv == -1)
+	if (gfrv != rv_ok)
 	{
 		return false;
```
skmp: This will always early exit in naomi builds

core/imgread/common.cpp
```diff
@@ -138,158 +152,122 @@ Disc* OpenDisc(const wchar* fn)
...
-	strncpy(fn,settings.imgread.LastImage, sizeof(fn));
-	fn[sizeof(fn) - 1] = '\0';
-
+	std::string diskImageFileName{ settings.imgread.LastImage };
 #ifdef BUILD_DREAMCAST
-	int gfrv=GetFile(fn,0,fileflags);
+	int gfrv = GetFile(diskImageFileName);
 #else
-	int gfrv=0;
+	int gfrv = rv_error;
```
skmp: Should probably be initialized to `rv_ok` here?

core/imgread/common.cpp
```diff
@@ -138,158 +152,122 @@ Disc* OpenDisc(const wchar* fn)
...
+    {
+        msgboxf("Selected image failed to load", MBX_ICONERROR);
+        NullDriveDiscType = Open;
+        gd_setdisc();
+        SetSnsCodes(NoDiskInsertedSenseCode);
+        return false;
+    }
+    else
+    {
+        SetSnsCodes(NoDiskErrorSenseCode);
```
skmp: Is this really correct here?

core/nullDC.cpp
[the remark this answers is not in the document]
skmp: I'd leave the #ifdef inside the if - only the file open logic is platform specific, not the empty string chech

core/windows/winmain.cpp
skmp: (i'm not pedantic about styling, so this looks okay to me. Would be best if changed)

skmp:
@pold500 did a review pass. What do you think about the comments?

core/imgread/common.cpp
pold500: Struct is the best option, imo.

core/hw/naomi/naomi_cart.cpp
pold500: lpstrFile is long pointer string, c_str() return constant char pointer.

core/imgread/common.cpp
pold500: kk... why braced-init is the problem?

core/imgread/common.cpp
pold500: I'm not sure. Why we should suppose everything is OK if we forgot to assign this variable a value, for example?

pold500: Also, tbh, I'm against the return codes. Makes code messy. We should use exceptions instead. I know we shouldn't use exceptions for controlling the code flow, but it seems to me nobody cares about return codes anyway, exceptions is harder to ignore.

core/imgread/common.cpp
pold500: I suggest I poorly chose a variable name. NoDiskErrorSenseCode - is that a "NoDisc!" error, or it is a "No Disc Error, everything's fine!".

core/imgread/common.cpp
pold500: Like, all we do with return code is just propagating it up and up the call stack until we can call "exit". Not much of a logic.

AbandonedCart: Harder to ignore, but also tend to cause more misdirected anger.

skmp:
This pull request **fixes 5 alerts** when merging 6fbf616489468f078bbc71a1585b9bf4e0cdd784 into 51a0c5727f2992004418d079631101da944eb917 - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-bebf6cd2ec7ae35301f3fa43b0a3365ce2c815ad)

**fixed alerts:**

* 5 for FIXME comment

---

core/imgread/common.cpp
baka0815: Because it adds clutter and is unnecessary here.

````c
if (...)
{
  ...
  return false;
}
return true;
````
would be enough.
The ``if`` always returns, so the ``else`` is unnecessary

File: core/nullDC.cpp (hunk 9 above)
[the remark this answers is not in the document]
Comment (baka0815) [reply core/nullDC.cpp]: @skmp there would be a nullpointer check which would do nothing then (everything in it would be inside the ``#ifdef``, so what's the point keeping it?

File: core/hw/naomi/naomi_cart.cpp (hunk 10 above)
Comment (skmp) [reply core/hw/naomi/naomi_cart.cpp]: `ofn.lpstrFile` is also used as output here? Is `SelectedFile` big enough?

File: core/imgread/common.cpp [hunk 16]
```diff
@@ -23,6 +23,24 @@ Disc*(*drivers[])(const wchar* path)=
...
 
+struct SenseErrorCode
+{
+    int sns_asc;
+    int sns_ascq;
+    int sns_key;
+};
+
+//No disc inserted at the time of power-on, reset or hard reset, or TOC cannot be read.
+const SenseErrorCode NoDiskInsertedSenseCode{ 0x29, 0, 6 };
```
skmp: why not init with `= { ... }`? Then it doesn't need C++11

skmp:
This pull request **fixes 5 alerts** when merging fa9e55d75f304265ade32bc707b60e0979953d4c into 14bf4b5b1c6a7dcbeb8ff83d47bb4f1fcf569773 - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-e56b2458891ef66dda69fcfefee39d6b9e86f7b2)

**fixed alerts:**

* 5 for FIXME comment

---

pold500:
@skmp so, what's up? is everything correct?

AbandonedCart:
Once https://github.com/reicast/reicast-emulator/pull/1275 is up, running, and merged, you should be able to rebase the PR on master to resolve the Travis issues and unlock merge.

AbandonedCart:
Performing a rebase to the current master should resolve the travis key issue and unlock merge.

AbandonedCart:
Is this still pending?

skmp:
@pold500 I'll pull this in as `pold500/fix-openfile-dialog` and do the rebase in a new PR

PR #1239, opened 2018-07-26
evdev: Allow two controls with the same event

### LINKED ISSUE #1042: Reicast crashes when two evdev_device_id_X entries match in emu.cfg
I'm attempting to get two players configured off of one input device (XArcade Tank Stick Dual if it matters), so my thought was to have two separate mappings of the device, one for each player's side, and then point them both to the same device id:

evdev_device_id_1 = 2
evdev_device_id_2 = 2
evdev_device_id_3 = -1
evdev_device_id_4 = -1
evdev_mapping_1 = /opt/retropie/configs/dreamcast/mappings/controller_XGaming XArcade.cfg
evdev_mapping_2 = /opt/retropie/configs/dreamcast/mappings/controller_XGamingXArcadeP2.cfg

np _(or nb, can't remember)_ = 2

This causes Reicast to immediately crash when opening anything. If I change one of the evdev id's, then it starts up, but the other mapping isn't used, and I can't control player 2.

Is this a bug, or am I missing something more obvious here?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47971205-reicast-crashes-when-two-evdev_device_id_x-entries-match-in-emu-cfg?utm_campaign=plugin&utm_content=tracker%2F500311&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F500311&utm_medium=issues&utm_source=github).
</bountysource-plugin>

baka0815:
My testing showed, that the check about the same event used in multiple configurations is no problem at all (tested with CAPCOM vs. SNK - Millennium Fight 2000 (PAL)).

Also the name of the configuration file was not copied correctly into the EvdevControllerMapping resulting in output like ``evdev: Using '}��U' mapping`` (instead of ``evdev: Using 'Logitech Logitech RumblePad 2 USB' mapping``).
Does my fix produce a memory leak? If so, where to free that?

Fixes #1042.

### DIFF
--- a/core/linux-dist/evdev.cpp
+++ b/core/linux-dist/evdev.cpp
@@ -158,8 +158,11 @@
 		ConfigFile mf;
 		mf.parse(fd);
 
+		string name = mf.get("emulator", "mapping_name", "<Unknown>");
+		char * cstr = new char[name.length()+1];
+  		strcpy (cstr, name.c_str());
 		EvdevControllerMapping mapping = {
-			mf.get("emulator", "mapping_name", "<Unknown>").c_str(),
+			cstr,
 			load_keycode(&mf, "dreamcast", "btn_a"),
 			load_keycode(&mf, "dreamcast", "btn_b"),
 			load_keycode(&mf, "dreamcast", "btn_c"),
@@ -192,6 +195,7 @@
 			mf.get_bool("compat", "axis_trigger_left_inverted", false),
 			mf.get_bool("compat", "axis_trigger_right_inverted", false)
 		};
+
 		return mapping;
 	}
 
--- a/core/linux-dist/main.cpp
+++ b/core/linux-dist/main.cpp
@@ -127,14 +127,6 @@ void SetupInput()
 			}
 			else
 			{
-				for (i = 0; i < port; i++)
-				{
-						if (evdev_device_id[port] == evdev_device_id[i])
-						{
-								die("You can't assign the same device to multiple ports!\n");
-						}
-				}
-
 				size_needed = snprintf(NULL, 0, EVDEV_DEVICE_STRING, evdev_device_id[port]) + 1;
 				evdev_device = (char*)malloc(size_needed);
 				sprintf(evdev_device, EVDEV_DEVICE_STRING, evdev_device_id[port]);


baka0815:
Again, travis failure unrelated.

core/linux-dist/evdev.cpp
```diff
@@ -158,8 +158,11 @@
 		ConfigFile mf;
 		mf.parse(fd);
 
+		string name = mf.get("emulator", "mapping_name", "<Unknown>");
+		char * cstr = new char[name.length()+1];
+  		strcpy (cstr, name.c_str());
 		EvdevControllerMapping mapping = {
-			mf.get("emulator", "mapping_name", "<Unknown>").c_str(),
+			cstr,
```
skmp: I think the right way to fix this is to use a `string` in `EvdevControllerMapping` instead of `char*`

skmp: https://github.com/reicast/reicast-emulator/blob/d8a7212d69929277d2c4ab54851fb7e97604ec3e/core/linux-dist/evdev.h#L7 make that `const string name`

skmp:
@baka0815 you are a colab. of the team now. You should be able to create namespaced branches and push to this repo directly. Please name your branches `baka0815/<descriptive-branch-name>` to make coordination with the rest of the team easier. 

PRs from within the repo run through more tests, and travis does not fail for them. We do have to fix travis at some point, but working from within the repo is a good enough workaround for now.

Thanks~

core/linux-dist/evdev.cpp
baka0815: I tried that first, didn't work however.
Might give that a second try, but I'm away for about a week now.

baka0815:
Ok, I'll keep that in mind for the next changes.

core/linux-dist/evdev.cpp
skmp: Any updates on this?

skmp:
(any updates on this?)

baka0815:
Infortinatly not, didn't have time to look into it and had hopes tha @TestSubject06 or @SteelBattalion could test my changes.
My current plan is to create a collab-branch to help that.

baka0815:
Closing for now, will reopen based on https://github.com/reicast/reicast-emulator/tree/baka/evdev_multiple

core/linux-dist/evdev.cpp
baka0815: I tested this again, and it's not working.

Changing the header to ``const string name;`` and ``load_mapping()`` to ``mf.get("emulator", "mapping_name", "<Unknown>"),`` (without ``.c_str()``) I get the following output:
``evdev: Using 'P�T%V' mapping``

So it's not working.
Also, changing the code to
````c++
		string name = mf.get("emulator", "mapping_name", "<Unknown>");
		EvdevControllerMapping mapping = {
			name,
````
doesn't work either.

PR #1257, opened 2018-08-05
Cleanup and forward-thinking integration

AbandonedCart:


shell/android-studio/reicast/src/main/java/com/reicast/emulator/emu/GL2JNIView.java
```diff
@@ -64,307 +65,308 @@
...
+        if(translucent) this.getHolder().setFormat(PixelFormat.TRANSLUCENT);
+
+        if (prefs.getBoolean(Config.pref_egl14, false)) {
+            setEGLContextFactory(new GLCFactory14.ContextFactory());
+            setEGLConfigChooser(
+                    translucent?
+                            new GLCFactory14.ConfigChooser(8, 8, 8, 8, depth, stencil)
+                            : new GLCFactory14.ConfigChooser(5, 6, 5, 0, depth, stencil)
+            );
+            GLES20.glEnable(GLES20.GL_DEPTH_TEST);
```
baka0815: As far as I can see, this is the only real functionality change, the rest is only whitespace changes.
I can't comment on this, as I don't know enough about GLES or Android.

PR #1271, opened 2018-08-09
Android: Fix the misleading "default disk" setting

AbandonedCart:
It is a boot disk, which would be the "default" disk. Labeling it default is misleading, though. It will also prevent confusion to have examples of common boot disks.

### DIFF
--- a/shell/android-studio/reicast/src/main/res/values/strings.xml
+++ b/shell/android-studio/reicast/src/main/res/values/strings.xml
@@ -52,7 +52,7 @@
     <string name="select_software">Force Software Rendering</string>
     <string name="select_sound">Disable Emulator Sound</string>
     <string name="select_depth">Rendering Depth</string>
-    <string name="default_disk">Default Disk</string>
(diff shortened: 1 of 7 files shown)


shell/android-studio/reicast/src/main/res/values/strings.xml
```diff
@@ -133,10 +133,6 @@
     <string name="downloadVMU">Download VMU</string>
     
     <string name="log_saved">Logcat saved and copied to clipboard\nPlease paste inside the Github issue</string>
-
-    <string name="service_ticker">reicast is running...</string>
-    <string name="service_title">reicast is running...</string>
-    <string name="service_content">Emulator service is currently running</string>
```
baka0815: Was the removing of those lines intentional?

AbandonedCart: https://github.com/reicast/reicast-emulator/commit/a9ea64f36a293f28507977ef403e4e9540cdc59c

baka0815: Ok, so they weren't used anymore.

PR #1280, opened 2018-08-11
Android: Add right stick switch to controller update

AbandonedCart:


AbandonedCart:
After verifying this, the default setting was *true* not *false*

PR #1273, opened 2018-08-10
evdev: Allow two controls with the same event

baka0815:
Includes the changes from #1239 but without the changes to ``EvdevControllerMapping`` to help test if this also works for more people than myself.

Fixes #1042 and #1035

PR #1283, opened 2018-08-12
X11: Values <= 0 disable keyboard

baka0815:
Value -1 for example doesn't disable X11 keyboard input, only 0 does (non zero = enabled).

Changed that behavior to match other settings (where -1 means disabled, ``joystick_device_id`` f.e.) so that every value above zero means enabled and zero or less means disabled.

Also print an information if the X11 keyboard is disabled (like with legacy joystick support).

I noticed this investigating #1273.

### DIFF
--- a/core/linux-dist/x11.cpp
+++ b/core/linux-dist/x11.cpp
@@ -180,7 +180,9 @@ void input_x11_init()
 	x11_keymap[KEY_F] = DC_AXIS_LT;
 	x11_keymap[KEY_V] = DC_AXIS_RT;
 	
-	x11_keyboard_input = cfgLoadInt("input", "enable_x11_keyboard", 1);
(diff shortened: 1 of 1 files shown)


AbandonedCart:
Looks good. If I could merge on a phone, I would. The failed build is an issue on their end.

baka0815:
Ok, then I'll go ahead and merge.

AbandonedCart:
It also didn't show that wercker was required on a phone. I assume that is blocking the merge, so I can fix that.

AbandonedCart:
If you repush the changes, it should restart the build and fix itself
```
git reset --soft HEAD^
git commit -m "X11: Values >= 1 enable keyboard"
git push -f
```

baka0815:
Yep, worked!

PR #1285, opened 2018-08-12
Travis: increase depth over fetching full repo

AbandonedCart:
Assuming the issue was that it could not access the "parent" commits, but any sliver of time that can be trimmed from the current build is desired.


AbandonedCart:
@skmp Does this still give you the same results?

AbandonedCart:
This worked after I increased the depth. Unshallow is just a less messy form of depth (really big number)

PR #1289, opened 2018-08-13
evdev/dupcheck: Don't check unassigned buttons

baka0815:
There is no need to check unassigned buttons (dpad2, D, Z f. e.) and this caused false positives.

Also removes the trailing white-spaces.

Follow up of #1277.


AbandonedCart:
Good addition.

PR #1298, opened 2018-08-18
Support per-game configuration files

AbandonedCart:
Each game could support it's own proprietary configuration that could be changed and updated, as needed. This separates per-game config from the emulator version and allows implementing "hacks" without making them actual hacks and avoids constantly monitoring compatibility.


AbandonedCart:
screenshot_20180820-120309

AbandonedCart:
https://github.com/reicast/reicast-emulator/blob/fh/mymaster/core/nullDC.cpp#L144 - 154:

```
[T30701D] // Pro Pinball Trilogy
Dynamic.safemode = 1

[T15112N] // Demolition Racer
Dynamic.safemode = 1

[T23001D] // Star Wars - Episode I - Racer (United Kingdom)
Dynamic.safemode = 1

[T7012D ] // Record of Lodoss War (EU)
Dynamic.safemode = 1

[T40218N] // Record of Lodoss War (USA)
Dynamic.safemode = 1
```

AbandonedCart:
https://github.com/reicast/reicast-emulator/blob/fh/mymaster/core/nullDC.cpp#L144 - 154:

```
[T30701D] // Pro Pinball Trilogy
Dynarec.safemode = 1

[T15112N] // Demolition Racer
Dynarec.safemode = 1

[T23001D] // Star Wars - Episode I - Racer (United Kingdom)
Dynarec.safemode = 1

[T7012D ] // Record of Lodoss War (EU)
Dynarec.safemode = 1

[T40218N] // Record of Lodoss War (USA)
Dynarec.safemode = 1
```

baka0815:
Maybe instead of comments in the INI file we should add a ``Name`` key without meaning?

````ini
[T30701D]
Name=Pro Pinball Trilogy
Dynamic.safemode = 1
````

**Comment by AbandonedCart** [comment]
Not sure what you mean. The comments are not part of the configuration file. There are there strictly as unread labels.

**Comment by AbandonedCart** [comment]
When used properly, you don't actually have to know the ID of the game. You can boot the emulator with a game, it will write the game settings to the config file, then you can modify it as desired. 
The emulator doesn't actually retrieve the game names, nor should that be documented as a value. That will only cause people to think it is required or that they have to find some specific value. It will lead to a lot of unnecessary documentation for something that is strictly for personal reference.

**Comment by baka0815** [comment]
I wouldn't want to check the name or something along those lines, just for information.
But we can also just use a comment, but then we should use INI comments?

````ini
[T30701D]
; Pro Pinball Trilogy
Dynamic.safemode = 1
````

**Comment by AbandonedCart** [comment]
It was literally just matching them for the Github comment. It's not part of the file.

**Comment by AbandonedCart** [comment]
The latest commit will print an arbitrary software name to the config, so it will appear as
```
[RDC-0140]
software.name = DEAD OR ALIVE 2 RDC
... (other settings) ...
```

AbandonedCart:
This is how mine looks after running two actual games
screenshot_20180820-132742

The only outstanding issue would be inconsistency between the version of the product_id using this code and the version obtained using the original code.

baka0815:
Hmm, I think we shouldn't use the emu.cfg for that, but use a ``games.cfg``.
One advantage of using a comment would be that it would be written first.

AbandonedCart:
All of the defaults are in the emu.cfg already, so it wouldn't make much sense to split them up. Also, what would be the benefit of having the user find the game ID and create a comment (that will be erased when the config loads) when it prints the official name?

AbandonedCart:
For those of you just tuning in, now a new game will save
```[reios_product_id]
software.name = reois_software_name```
to the emu.cfg and allow custom configuration for any values you want to *differ* from the base configuration.

baka0815:
Ok, fair enough!

AbandonedCart:
The hideous redundancy was purely a "hey, it works" lol

AbandonedCart:
Perhaps the most beautiful fragment ever seen.... This deserves a bounty...

AbandonedCart:
RELEASE THE KRAKEN!

Issue #1240, opened 2018-07-28
Whitespace consistency


LoungeKatt:
After reviewing the changes for https://github.com/reicast/reicast-emulator/pull/1238 and seeing whitespace issues cited against existing whitespace inconsistencies, my questions is can we pick a specific format, clean up the code, and then follow it?

tabs / 4 spaces?
separated / condensed statements?
new line / same line brackets?

pold500:
I also propose installing clang format plugin into build tools and writing a format.clang file with the rules, so they would be enforced.

baka0815:
I personally would indent with tabs.
Regarding brackets I don't have a strong opinion.

pold500:
I have no strong opinion about what exactly the style should be too, but I have the strong opinion that it should be enforced in the project, both when you write code, and in build verification tools, like CI.

I propose:
1. For variable names - underscore style (this_is_variable)
2. if/else/statements - 
if( true )
{
}
else
{
//always a brackets even for one statement here
}
3. spaces - spaces instead tabs, 4 space for a tab.

LoungeKatt:
@pold500 Android Studio uses camel case variables by default. It's probably best not to create some arbitrary rule that involves needlessly submitting 10,000 lines of edits to "change some names" that git has no concern over, while whitespace issues do cause merge fails.

It also happens to cite "simplifying statements" as a warning and recommends removing unnecessary brackets, if not creating the `variable ? true ; false` shorthand. This was more a question of what should be the general rule in terms of spacing. I don't think restricting the entire design of the statement to one format is rational.

baka0815:
I would prefer
````C
if (condition)
{
<tab>something
}
else
{
<tab>something
}
````

Use brackets whenever there is more than one line in at least one block. If all cases consist of just one line then no brackets.

Regarding variable names I'm fine with either ``variable_name`` (snake case) or ``variableName`` (canel case).

LoungeKatt:
@baka0815 Seems reasonable. That is something both Xcode and Android Studio could handle (which are my primary concern because Windows / Linux are pretty adaptable)

skmp:
I have a stylesheet somewhere from the nulldc/e days.

in general, 4-spaces ident, and the C# styling guide.

skmp:
Ideally, we'd use clang-format and also have validation for it - it has worked great for other projects.

pold500:
Here's an example of .clang-format file.

We can discuss every rule, how we want it to be.

---
[...]

PR #1311, opened 2018-08-21
RFC: Exemplary .clang-format file

baka0815:
Very basic clang-format configuration file.

Indentation is via Tabs which are 4 columns wide.
Brackets are in the next line.

Example:
````cpp
if (condition)
{
    something();
}
````

However I'm not sure about the precompiler settings, whichbare currently "no indentation":
````cpp
#if DEFINE
#if SUBDEFINE
#endif
#else
#endif
````

I would have liked
````cpp
#if DEFINE
    #if SUBDEFINE
    #endif
#else
#endif
````

But the only setting I found would have made the code look like
````cpp
#if DEFINE
#    if SUBDEFINE
#    endif
#else
#endif
````

See #1240.

### DIFF
--- a/core/.clang-format
+++ b/core/.clang-format
@@ -1,0 +1,27 @@
+IndentCaseLabels: true
+IndentPPDirectives: None
+IndentWidth: 4
+TabWidth: 4
+UseTab: Always
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortBlocksOnASingleLine: false
+PointerAlignment: Left
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: true
+  AfterControlStatement: true
+  AfterEnum: true
+  AfterFunction: true
+  AfterNamespace: true
+  AfterStruct: true
+  AfterUnion: true
+  AfterExternBlock: true
+  BeforeCatch: true
+  BeforeElse: true
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+SortIncludes: false


AbandonedCart:
Many compilers require that preprocessor definitions are not preceded by whitespace. Chances are you didn't find the format you wanted because it's not widely supported.

baka0815:
Maybe, but what I cited as what I "would have liked" is what we're currently doing.
So we need to talk about that as it's not possible voa clang-format (or I did not find it).

AbandonedCart:
Well then you have three options:
1. enforce the "unfavorable" way via .clang-format
2. document the preference externally and police it manually
3. do nothing in hopes people continue to follow the pattern

baka0815:
Unfortunately not. :-(

I just tested removing ``IndentPPDirectives=None`` from the ``.clang-format`` file and the result is the same.
Example using ``core/stdclass.cpp``:
````patch
-#if BUILD_COMPILER==COMPILER_VC
-       #include <io.h>
-       #define access _access
-       #define R_OK   4
+#if BUILD_COMPILER == COMPILER_VC
+#include <io.h>
+#define access _access
+#define R_OK 4
 #else
-       #include <unistd.h>
+#include <unistd.h>
 #endif
````

AbandonedCart:
So there are still 3 options to choose from:
1. enforce the "unfavorable" way via .clang-format
2. document the preference externally and police it manually
3. do nothing in hopes people continue to follow the pattern

baka0815:
But there is no way to "police it manually" (or I'm not getting what you mean) because clang-format would change it to one way ore the other.

AbandonedCart:
Policing manually is more or less what you do now (line comments, reviews, etc)

dmiller423:
Well i've never seen a compiler that wouldn't take spaces after # (on defines esp.).
Either way,  I don't see a compiler specific tool to format the source as a great idea myself.
This is also not really a pull request,  you mean to include this in make? cmake? ide? which?

AbandonedCart:
@dmiller423 Spaces **before** the #

baka0815:
@dmiller423 see #1240 where we're talking about consistency in the (core) code.
We could include the ``.clang-format`` file in the pr-checks to see if the changes comply with our styleguide.
Including the file in the code base makes it also easier for everyone to format your code using ``clang-format <source file>``.

@AbandonedCart ok, now I get what you mean. You're talking about not enforcing anything via automated formating and just commenting during reviews to keep the style clean.

AbandonedCart:
It's not a glorious option, but it's still an option.

baka0815:
Closing this for now.

skmp:
Will complete the review after the Orthodox Easter

skmp:
(@baka0815 I'm also gonna close this one, and we can revisit after the refactor)

PR #1312, opened 2018-08-21
Android: Add support for dynamic peripherals

AbandonedCart:


shell/android-studio/reicast/src/main/jni/src/Android.cpp
```diff
@@ -265,7 +292,11 @@ void os_SetupInput()
 	for (int i = 0; i < 3; i++)
 	{
 		if (add_controllers[i])
-			mcfg_CreateController(i+1, MDT_None, MDT_None);
+			mcfg_CreateController(
+                    i + 1,
+                    GetMapleDeviceType(controller_periphs[i][0], i + 1),
```
AbandonedCart: Yes, I know these are one off

baka0815: Yeah, because of add_controllers starting wirh player 2. :)

AbandonedCart:
*first steps to dynamic peripheral support*
Breathe. It's not being merged as is. Holy %&#!...

baka0815:
I left some comments on the commit.
Sorry, for messing that up...

AbandonedCart:
Without a connected UI, you are safe to assume it isn't done

PR #1314, opened 2018-08-22
Android: General updates and improvements

AbandonedCart:


.travis.yml
```diff
@@ -44,7 +44,7 @@ install:
 - test -z "$encrypted_c726d225a9d9_key" || mv debug.keystore ~/.android/debug.keystore
 before_script: 
 script:
-- git fetch --depth 2
```
skmp: I think I do a full depth somewhere? It is needed to get the tags / versioning right. Was that changed or is that unrelated?

AbandonedCart: Depth is now 50 on travis to reduce the size and sync time without impact on results.

PR #1316, opened 2018-08-22
Android: Configure peripherals independently

AbandonedCart:


baka0815:
What's the purpose of this? Adding peripherals without (additional) controllers?

AbandonedCart:
or to configure peripherals independently, whichever wording you like better.

https://github.com/reicast/reicast-emulator/issues/1208#issuecomment-415161153

baka0815:
Because it shall be possible to enable 2 VMUs and a microphone with just 1 controller?

baka0815:
Well it's not actual Dreamcast behavior but it's fine for Android I think.

AbandonedCart:
@baka0815 reicast has forced using controller 2, port 1 as the microphone for 3 years. I expanded that so that you could now pick any of three controllers to have it plugged into. If you prefer, I can restore the original configuration and go back to locking out that port.

AbandonedCart:
Not a whole lot of options here. 

This commit fixes Seamen, but requires the user pick a port in controller 2 - 4 for the mic. 
Reverting to the old method fixes it at the cost of hardcoding a port to act as the mic.
Adding controller 1 as an option requires unlocking it to the risk of data loss, misconfiguration, user error, and a list of other possible issues.
Doing nothing makes Seamen unavailable without a hardware controller in port 2.

Take your pick. What gets screwed?

AbandonedCart:
Well now it will force controller 1, port 2 and *then* allow the use of any additional microphones (although I have no idea why you would have more than 1, but people deserve options)

blackman91:
Does that mean we can no longer access vmu_save_A2.bin and that the microphone will always be enabled? Also does that mean the enable microphone(s) option in settings option is obsolete now? Should it be removed?

AbandonedCart:
No and no.

AbandonedCart:
The mic toggle is as it was before multiple controllers, but now enables any additional user-configured microphones along the ay.

PR #1321, opened 2018-08-24
Add support for libao 

gameblabla:
This adds libao support to reicast. Now, you may ask why even bother when reicast already has ALSA, OSS, Pulseaudio, Coreaudio etc... support ?
Well for once, libao sounds much better than the ALSA code especially when things get slow... (When things get slow, it constantly make a screeching sound over ALSA and constantly complains about underruns)
While it's not perfect in this case, it's still much more pleasant experience for me, at least on Void linux Musl...
It also supports for the other sound APIs supported by libao. 
As you can see, the code for it is fairly trivial.
To avoid a dependency on libao, i've disabled it by default.


AbandonedCart:
What about all the other platforms besides Linux?

gameblabla:
Well i admit i haven't tested it on other platforms like Windows or Mac OSX. In theory, libao is supported on those and more such as OpenBSD, IRIX, Solaris... 
However, it might not be practical to try reicast on all of these platforms due to other issues... (namely the SDL2 port not working at all right now, i'll open another issue about it)

While the code is fairly small and trivial, i also don't know if it works ok on Windows or Mac... Sorry, i can't try reicast on those D :

AbandonedCart:
@baka0815 It's one thing to fine tune something through additional PRs later, but you are talking about not even verifying it works. With everything that already needs to be completed, do we really want to start adding partially implemented drivers, too?

dmiller423:
It's optional,  and if it works in linux it's highly likely that it works just as well anywhere else it's avail.
I do agree perhaps someone should test it in windows and osx,  mostly to be sure it's added to the build system there optionally as well...   Though this is likely a reason many people don't want to add to projects,  not everyone wants to deal with windows or has a machine running osx.

AbandonedCart:
@dmiller423 Given your recent experiences, you should know that isn't necessarily a true statement.

That aside, it isn't that hard to ifdef it out to just the platforms it is verified on as a reminder that it is not yet available for others. It's not as though you necessarily have to implement it on all of them, but then you can't leave it partially implemented and untested. That is how people end up enabling a build flag for something and then submitting issues about it being broken when it was never intended to be enabled in the first place.

gameblabla:
Preferably, the libao code would need to be tested at least on Windows & Mac OS X. But even if i add to put ifdefs and conditions for libao, where would i put them ? There's no clear way to do so. OSS isn't available on Windows and despite this, it's still possible to turn on the switch and attempt to compile. Of course, it will fail to compile. So yeah; i don't think conditions will be very much useful in this case.

core/oslib/audiostream.cpp
```diff
@@ -9,6 +9,7 @@
 #include "oslib/audiobackend_pulseaudio.h"
 #include "oslib/audiobackend_coreaudio.h"
 #include "oslib/audiobackend_omx.h"
+#include "oslib/audiobackend_libao.h"
```
AbandonedCart: Surround this with `#if USE_LIBAO`

AbandonedCart:
If @dmiller423 and @baka0815 want to assume responsibility for testing other platforms, then I won't worry about whether or not it actually works.

gameblabla:
Alright, i've removed the printf just in case the platform in question does not have printf available. I've also added if (aodevice) to ao_play also in case that it does not get initialized properly. Maybe that should be better ?

AbandonedCart:
That is enough to prevent a meltdown, so that's fine.

core/oslib/audiobackend_libao.cpp
```diff
@@ -0,0 +1,50 @@
...
+	memset(&aoformat, 0, sizeof(aoformat));
+	
+	aoformat.bits = 16;
+	aoformat.channels = 2;
+	aoformat.rate = 44100;
+	aoformat.byte_format = AO_FMT_LITTLE;
+	
+	aodevice = ao_open_live(ao_default_driver_id(), &aoformat, NULL); // Live output
+	if (!aodevice)
+	{
```
baka0815: Please remove the brackets for single line indentation.

core/oslib/audiobackend_libao.cpp
```diff
@@ -0,0 +1,50 @@
...
+		aodevice = ao_open_live(ao_driver_id("null"), &aoformat, NULL);
+	}
+}
+
+static u32 libao_push(void* frame, u32 samples, bool wait)
+{
+	if (aodevice) ao_play(aodevice, (char*)frame, samples * 4);
```
baka0815: Please break after the condition:
````cpp
    if (aodevice)
        ao_play(aodevice, (char*)frame, samples * 4);
````

File: core/oslib/audiobackend_libao.cpp [hunk 4]
```diff
@@ -0,0 +1,50 @@
...
+	
+	return 1;
+}
+
+static void libao_term() 
+{
+	if (aodevice)
+	{
```
baka0815: Same as above.

baka0815:
I can test on Windows (maybe monday), however I can't test on Mac.

core/oslib/audiobackend_libao.cpp
gameblabla: I don't understand what i should do here ? i can't remove the brackets since there are two functions inside.

AbandonedCart: Yeah, the one is fine, as is.

skmp:
I think we can merge this as is, and do a follow up for windows / osx support. How does that sound?

baka0815:
Sorry, I can't do the testing on Windows as VirtualBox only supports OpenGL 2.1 and doesn't support wglCreateContextAttribsARB.

gameblabla:
@baka0815 If you could make a build for others to try though, that would be nice. Make sure the other audio apis are disabled of course...

baka0815:
Yes, but not working on wine tells you nothing. I may take a look at this again later this week.

baka0815:
I can confirm it's working on Windows (thanks to the software renderer).

However the DLL is linked statically so a build against libao won't work without it.
That should be changed as libao is not that much widespread under Windows (it took hours to find a precompiled DLL).

PR #1327, opened 2018-08-25
WIP: Fixes for Musl

gameblabla:
This allows it to compile & run on Void Linux musl (and possibly other musl-based linux distribution such as Alpine Linux) either via cmake or the makefile.
The file sys/cdefs.h in musl does not exist (see https://wiki.musl-libc.org/faq.html, Q: When compiling something against musl, I get error messages about sys/cdefs.h) so it needs to be removed.
#include <unistd.h> does not exist on Void linux musl either and even on glibc, GCC recommends that we use unistd.h instead rather than sys/unistd.h. (at least on recent versions of GCC on a modern distribution that is...)

There's also this line of code in core/deps/libwebsocket/sha-1.c
static void sha1_step __P((struct sha1_ctxt *));

Upstream does not have this line of code in question and it builds fine without it.

It's possible that removing <sys/cdefs.h> could prevent it to compile on other platforms. 
It compiles and works on both Devuan (Glibc) & Void linux musl but pls test thx on other platforms and i hope i can gather feedback on how to properly fix this.

### DIFF
--- a/core/deps/libwebsocket/sha-1.c
+++ b/core/deps/libwebsocket/sha-1.c
@@ -99,8 +99,6 @@ static const unsigned int _K[] =
 		sha1_step(ctxt);		\
 	}
 
-static void sha1_step __P((struct sha1_ctxt *));
-
 static void
 sha1_step(struct sha1_ctxt *ctxt)
 {
--- a/core/deps/ifaddrs/ifaddrs.h
+++ b/core/deps/ifaddrs/ifaddrs.h
@@ -44,11 +44,13 @@ struct ifaddrs {
 #define	ifa_broadaddr	ifa_dstaddr	/* broadcast address interface */
 #endif
 
-#include <sys/cdefs.h>
-
-__BEGIN_DECLS
+#ifdef _cplusplus 
+extern "C" {
+#endif
 extern int getifaddrs(struct ifaddrs **ifap);
 extern void freeifaddrs(struct ifaddrs *ifa);
-__END_DECLS
+#ifdef _cplusplus 
+}
+#endif
 
 #endif
--- a/core/deps/libwebsocket/private-libwebsockets.h
+++ b/core/deps/libwebsocket/private-libwebsockets.h
@@ -203,7 +203,6 @@ typedef unsigned __int64 u_int64_t;
 #else
 
 #include <sys/stat.h>
-#include <sys/cdefs.h>
 #include <sys/time.h>
 
 #if defined(__APPLE__)
--- a/core/oslib/audiobackend_oss.cpp
+++ b/core/oslib/audiobackend_oss.cpp
@@ -2,11 +2,7 @@
 #ifdef USE_OSS
 #include <sys/ioctl.h>
 #include <sys/fcntl.h>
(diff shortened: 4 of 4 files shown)


AbandonedCart:
After posting the first two, it may save time to simply write it here:
Why not make a target and add / edit conditionals instead of hardcoding changes for one platform that will be applied to all of them?

gameblabla:
> Why not make a target and add / edit conditionals instead of hardcoding changes for one platform that will be applied to all of them?
I haven't done it in this case because musl libc cannot be reliably detected and it is strongly discouraged by the authors themselves. It also works on glibc as well as musl and i believe my approach is the best and will avoid cluttering the code with un-necesarry lines of code that will not solve anything... 
Given that the code itself should in theory work on other *nix platforms, i believe it is not required.
I would have done it though if i had to work on something like evdev though.

AbandonedCart:
You mean to say building does not use a makefile for your platform?

gameblabla:
> You mean to say building does not use a makefile for your platform?
Makefiles, Cmake files and all... are still the same for musl & glibc. If you look at the commits, it's only a few lines of changes and mostly conformance issues. (musl is more strict than glibc)

AbandonedCart:
Keep in mind, you are editing files that are used by Mac, Windows, Android, iOS, and a few others.

You don't appear to have a whole lot of confidence in whether or not your changes are destructive, so adding conditions seems far safer than leaving it to chance. What I was hinting at is that having the same makefiles means you also have the same capability to add a manually set flag that could enable or act as the target definition. There is no reason that detection of the platform should even matter.

gameblabla:
> Why not TARGET_MUSL not unlike those for nearly every other platform included in the source?

That could work too but i would rather not to, sorry. I believe it should be treated like glibc. (besides, sys/cdefs.h is a glibc header that is not supposed to be exposed to applications)
If the commit as it is fails to convince you guys, then we are at a standstill.

AbandonedCart:
I'm sure you understand that "I would rather not" is not a valid enough reason to simply accept making destructive changes to the core source.

gameblabla:
Then i guess i'll wait for someone to try to compile it on Windows, Mac and stuff... So far, the automatic checks have passed in addition to it working on both Devuan (glibc) and Void linux musl (which i have personally tried).

skmp:
(renamed to use the WIP prefix, is this a WIP or an RFC?)

core/deps/libwebsocket/sha-1.c
```diff
@@ -99,8 +99,6 @@ static const unsigned int _K[] =
 		sha1_step(ctxt);		\
 	}
 
-static void sha1_step __P((struct sha1_ctxt *));
```
skmp: This should be `static void sha1_step (struct sha1_ctxt *);`, as __P is part of cdefs. Seems to compile fine without the declaration though?

skmp:
The changes look fine to me (with my limited POSIX knowledge and a bit of googling). Though, I do generally agree with @AbandonedCart that it is best to hide things behind an `#ifdef` if you're not willing to do due diligence on how other platforms are affected :)

PR #1328, opened 2018-08-25
vs: Upgrade vcxproj to visual studio 2015

AbandonedCart:
See https://github.com/reicast/reicast-emulator/commit/3579c9aabf673c241253ce184f394a5ec293d848

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,15 @@
 .DS_Store
 *.class
 *.orig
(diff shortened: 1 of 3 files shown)


AbandonedCart:
Note: These are the automatic changes that Visual Studio performs when opening the project using Visual Studio 2015 (Enterprise)

PR #1329, opened 2018-08-25
Mac: Reference "official" preprocessor directives

AbandonedCart:
This allows checking against the whole library of definitions specifically for Mac, including the wondrous `TARGET_OS_MAC` that checks for anything Apple-based.

### DIFF
--- a/core/linux/common.cpp
+++ b/core/linux/common.cpp
@@ -5,6 +5,7 @@
 #if HOST_OS == OS_DARWIN
 	#define _XOPEN_SOURCE 1
 	#define __USE_GNU 1
+	#include <TargetConditionals.h>
 #endif
 #if !defined(TARGET_NACL32)
 #include <poll.h>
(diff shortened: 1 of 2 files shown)


core/linux/common.cpp
```diff
@@ -294,7 +295,7 @@ void enable_runfast()
 }
 
 void linux_fix_personality() {
-        #if !defined(TARGET_BSD) && !defined(_ANDROID) && !defined(TARGET_OS_IPHONE) && !defined(TARGET_NACL32) && !defined(TARGET_EMSCRIPTEN)
```
skmp: `TARGET_OS_MAC` is used now instead of `TARGET_OS_IPHONE`?

AbandonedCart: No, `TARGET_OS_MAC` applies to OSX / iOS, while `TARGET_OS_IPHONE` is exclusively for iOS.

PR #1334, opened 2018-08-26
Extend per-game configuration options

AbandonedCart:


core/nullDC.cpp
```diff
@@ -129,15 +129,26 @@ void* webui_th(void* p)
 cThread webui_thd(&webui_th,0);
 #endif
 
+#if defined(_ANDROID)
+int reios_init_value;
```
baka0815: Why introduce this here?
Can't you do ``reios_init_value = dc_init();`` from Android?

AbandonedCart:
While not every regular setting has been made available, this removes the block on adding all but reios itself, since that would require a duplicate init and should not be to inconvenient to toggle universally.

core/nullDC.cpp
AbandonedCart: No. That defeats the purpose.

AbandonedCart: The first half of the init is required to obtain valid value for the disk info, but also contains the return values (in the regular init). The second half of the init would not run if a return value is set, but needs to be run after the disk info.
If you do not store the values of the first half, the second half will run unchecked. If you run the second half before the values are obtained, none of the settings are applied.

baka0815: And why do this only on Android?
Wouldn't that be good to do on other platforms too?

AbandonedCart: It is on other platforms.

AbandonedCart:
screenshot_20180826-115319

🥇

AbandonedCart:
... definitely shouldn't be taking Travis 13 minutes to mark a successful build successful...

PR #1339, opened 2018-08-27
evdev: Search for libevdev.so.2 first

### LINKED ISSUE #1336: Missing symlink for libevdev.so can result in failed auto-detection
**Please Note: This form is the minimum required information for submitting bugs.**  
**Removing this form may lead to your issue being closed until it is completed.**

| Platform | Branch | Hash | CIDL |
| ------ | ------ | ---- | ---- |
| Void Linux Musl | master | dunno | From source |

**_Description of the Issue_**

The evdev code expects the dynamic library to be named "libevdev.so".
https://github.com/reicast/reicast-emulator/blob/master/core/linux-dist/evdev.cpp#L29
However on some linux distributions (Void linux, Debian so far), such symlink does no exist for evdev, so it will fail the autodetect on those unless you create a symlink or modify the source code.

**_Debugging Steps Tested_**

  * Creating a symlink (seems to work)
  * Change "libevdev.so" to "libevdev.so.2"

**_Logs Gathered_**

```

```

**_Screenshots_**

None needed

- [X] Place an X in the box to confirm you have filled in this form

baka0815:
Many Linux distributions provide livevdev.so.2 only so check that first.

See https://packages.debian.org/sid/amd64/libevdev2/filelist for Debian. You would need to create a symlink yourself or install ``libevdev-dev`` (see https://packages.debian.org/sid/amd64/libevdev-dev/filelist) which no normal user should have installed as it's a developer package.

Fixes #1336

### DIFF
--- a/core/linux-dist/evdev.cpp
+++ b/core/linux-dist/evdev.cpp
@@ -21,12 +21,12 @@
 	void load_libevdev()
 	{
 		if (libevdev_tried)
-		{
 			return;
-		}
 
 		libevdev_tried = true;
-		void* lib_handle = dlopen("libevdev.so", RTLD_NOW);
+		void* lib_handle = dlopen("libevdev.so.2", RTLD_NOW);
+		if (!lib_handle) // libevdev.so.2 not found, fallback to libevdev.so
+			lib_handle = dlopen("libevdev.so", RTLD_NOW);
 
 		bool failed = false;
 
@@ -56,7 +56,7 @@
 			}
 		}
 
-		if(failed)
+		if (failed)
 		{
 			puts("WARNING: libevdev is not available. You'll not be able to use button names instead of numeric codes in your controller mappings!\n");
 			return;


core/linux-dist/evdev.cpp
```diff
@@ -21,12 +21,12 @@
...
 		if (libevdev_tried)
-		{
 			return;
-		}
 
 		libevdev_tried = true;
-		void* lib_handle = dlopen("libevdev.so", RTLD_NOW);
+		void* lib_handle = dlopen("libevdev.so.2", RTLD_NOW);
+		if (!lib_handle) // libevdev.so.2 not found, fallback to libevdev.so
+			lib_handle = dlopen("libevdev.so", RTLD_NOW);
```
AbandonedCart:
```
if (!dlopen("libevdev.so.2", RTLD_NOW))
    dlopen("libevdev.so", RTLD_NOW);
```
Wouldn't work?

AbandonedCart:
Am I missing something, or was all of the trouble really for two references?

AbandonedCart:
I get why it failed, but not why it was too much trouble to convert away from using it.

PR #1348, opened 2018-08-27
Begin implementing disk swap options

AbandonedCart:


AbandonedCart:
The testing branch for this PR on builds.reicast.com is "lk/disk-swap"

AbandonedCart:
This works as is for using a Utopia disk, Gameshark, etc.
For games like Shenmue that have multiple disks, it would be a bit irrational to define one of the disk (assuming they are individual images) to a hardcoded boot disk. A secondary option to select a disk at runtime is needed.
For games that always require Utopia, this could be added as a PGC option.

AbandonedCart:
Without having a Gameshark Lite (the version that did not require a special VMU), this can still be tested using two regular games.
Set "Game 1" as the boot disk.
Select "Game 2" to launch. Do nothing and verify that Game 1 is launched.
Return to the game selection and select Game 2 again.
This time, before the initial BIOS animation completes, open the onscreen menu.
Click "Disk Swap" and verify that Game 2 is loaded.

AbandonedCart:
Broken by https://github.com/reicast/reicast-emulator/pull/1238/commits/82c0ef59f81c60fb81b898057cce6623d9827eb0

AbandonedCart:
Moved to https://github.com/reicast/reicast-emulator/pull/1356 due to issues with PR updating

PR #1356, opened 2018-08-28
Android: Support disk swapping / boot disk

AbandonedCart:


AbandonedCart:
https://github.com/reicast/reicast-emulator/pull/1358 will break with this, but with minimal required change. Rather than converting the `jstring` to `const char*`, it would simply be converted to `std::string` for the `bootdisk` and `diskSwap` methods. This portion of the code is known to have no issues with performing the conversion.

PR #1367, opened 2018-09-02
add save states

RetroSven:
add save states

currently only has input hooks for Linux (f2 save, f4 load) but should be trivial for the main developers of the other platforms to add them 

may need to add some clock/time include/ifdefs for the other platforms to make get_time_usec function properly, but that should also be trivial


.gitignore
```diff
@@ -35,6 +35,11 @@ reicast-ios.xccheckout
 shell/linux/.map
 shell/linux/nosym-reicast.elf
 shell/linux/reicast.elf
+.map
+nosym-reicast.elf
+reicast.elf
+Makefile
```
baka0815: We shouldn't ignore Makefiles.

RetroSven: The Makefile ignore is essentially the same as the pre-existing Makefile ignore for shell/linux/Makefile.  I copied that Makefile into the root dir so that Eclipse would pick it up correctly.

I'm not clear on what you're looking for wrt commits.  The vast majority of all the changes to the "core" files are simply to move the structs in cpp files to their corresponding .h file (or types.h) so that they can be referenced from serialize.cpp.  

Some new methods were created from the removal of the corresponding methods inside of structs.  E.g.
struct foo {
    u32 var ;
    void method() ;
}
becomes
struct foo {
    u32 var ;
}
void foo_method() ;

There really isn't much more to it than that as far as changes to "core"

baka0815: Yeah, since most of the line count comes from the moving of the cose to the header files I would like to have them in a saparate commit to make the review process much easier for the real changes.

blackman91:
That's awesome, I thought the save states were only possible on the libretro core.

.gitignore
AbandonedCart: You could also replace the existing ignores with the new ones. `shell/linux/X.Z` is a specific instance of the file, but the general `X.Z` will also ignore it.

RetroSven: btw unfortunately I won't have time to split this commit into two separate one (if anyone is waiting for that to happen), but if you have any questions about any of this please let me know.  This commit was mainly to grant permission to use the code that I originally developed for the libretro fork.  (I saw someone made a comment on one of the save state issues that you couldn't just grab the libretro code without permission)

AbandonedCart: I was saying you'd need the author to sign the CLA, which you did. Thanks.

AbandonedCart:
I get the same `undefined reference to 'decoded_colors'` locally. I am assuming something was changed in the original base that made this available to the code, which will need to be addressed when this is packed up into a merge-ready PR. Thanks for making this available.

AbandonedCart:
Also, it would appear that `aica_pending_dma` was created by the changes that were removed in https://github.com/reicast/reicast-emulator/commit/ee3a474ee45dafaed40fadcc415ed418341fb604

RetroSven:
decoded_colors is in softrend.cpp:

u32 decoded_colors[3][65536];

so wherever you're getting that undefined reference, I think you just need 

extern u32 decoded_colors[3][65536];

Odd that you're getting that - everything compiled clean on my end before I made the original PR.

AbandonedCart:
It's getting it from serialize in both appveyor and wercker

```
/pipeline/source/shell/linux/../../core/./serialize.cpp:820: undefined reference to `decoded_colors'
```

RetroSven:
I suppose there's an ifdef somewhere that does not include softrend.cpp for those builders

AbandonedCart:
Possibly. I'm not overly fond of all of the extern usage (by not fault of yours) because it's more or less just "promising the compiler a definition will exist" to quote someone's clever line on SO. There's no solid links between files, but trying to put them in there causes an even bigger issue.

RetroSven:
surround the extern decoded_colors in serialize.cpp with

			#if FEAT_HAS_SOFTREND
should fix that

AbandonedCart:
🤦‍♀️ Why didn't I think of that? lol, thanks.

baka0815:
> I suppose there's an ifdef somewhere that does not include softrend.cpp for those builders

Yes, we have some builds that don't support software rendering.

PR #1368, opened 2018-09-03
Fix 32-bit vram bank mapping

flyinghead:
Fixes texture corruption and other problems in many games

### DIFF
--- a/core/hw/pvr/pvr_mem.cpp
+++ b/core/hw/pvr/pvr_mem.cpp
@@ -281,14 +281,14 @@ u32 pvr_map32(u32 offset32)
 {
 		//64b wide bus is achieved by interleaving the banks every 32 bits
 		const u32 bank_bit = VRAM_BANK_BIT;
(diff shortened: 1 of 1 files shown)


baka0815:
Could you provide screenshots (before/after) or specify where I should look to notice the changes?

flyinghead:
Rayman 2

Before:
rayman2-before

After:
rayman2-after

flyinghead:
Re-Volt
Before:
re-volt-before

After:
re-volt-after

skmp:
@flyinghead would be nice in the future to have some rationale in the PR, apart from the fixes, explaining why you think this is a more correct approach. Otherwise, we could be fixing other bugs by introducing more bugs :)

flyinghead:
The shifting of the offset part was wrong (3 instead of 1). I changed it back to what is was before this function was changed: https://github.com/reicast/reicast-emulator/commit/f571d6c594fc9efd11d439c3a38afd401031e9d0

AbandonedCart:
(unless the commit message is explicit enough to make it redundant)

PR #1373, opened 2018-09-04
Softrend: works only with X11

baka0815:
If X11 support is not there, reicast doesn't compile because the softrenderer needs it.
Use #ifdef to do X11 only if SUPPORT_X11 is set.

### DIFF
--- a/core/rend/soft/softrend.cpp
+++ b/core/rend/soft/softrend.cpp
@@ -1226,8 +1226,10 @@ struct softrend : Renderer
 		XPutImage(x11_disp, x11_win, gc, ximage, 0, 0, (x11_width - width)/2, (x11_height - height)/2, width, height);
 		XFree(ximage);
 		XFreeGC(x11_disp, gc);
+	#else
+		// TODO softrend without X11 (SDL f.e.)
+		#error Cannot use softrend without X11
 	#endif
-	// TODO softrend without X11
 #endif
 	}
 };
--- a/core/rend/soft/softrend.cpp
+++ b/core/rend/soft/softrend.cpp
@@ -8,7 +8,7 @@
 
 	Initial code by skmp and gigaherz
 
-	This is a rather weird very basic pvr softrend. 
+	This is a rather weird very basic pvr softrend.
 	Renders	in some kind of tile format (that I forget now),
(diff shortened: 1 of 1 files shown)


dmiller423:
yes I believe the only support w.o X11 is GLES on EGL

dmiller423 (commented):
You might consider using #elif instead of another ifdef as well as another else afterwards w. a #error since you're effectively turning it into not rendering anything...  Is this even useful?   we could just do a 
#if defined(NO_X11) && defined(SOFTREND)  
#error can't use softrend w.o X11   
#endif

baka0815:
We could do that, however we would also need to disable SOFTREND on SDL then.

baka0815:
After confirming that GLES still renders (however without GUI on Linux) and SDL still working, I went for the ``#else`` route because we might add additional software outputs (like SDL) in the future.

dmiller423 (changes requested):
1211: #elif defined(SUPPORT_X11)
1212: delete 
1229: remove indent
1232: delete

dmiller423:
SDL w.o X11?   what's it use directfb ?

baka0815:
SDL uses GLES or OpenGL:
https://github.com/reicast/reicast-emulator/blob/1c880fc37c64b3dd1d79df1f5d90275ed1a6ef02/core/sdl/sdl.cpp#L463

dmiller423:
Without X11?

baka0815:
At least without the X11 define, yes. SDL should also work (theoretically) on wayland.
If you build SDL with X11, you'll get two windows (X11 and SDL) however only the SDL one renders.

dmiller423 (approved):

dmiller423:
lmao,  the define is to control whether or not to use X11 directly ... the question refers to whether or not SDL currently works on linux w.o any X11/Xorg ( aka direct from console ) ...   
Either way the changes amount to nothing more than a #ifdef/error, albeit one more informative than the other error message...  I will merge it and close.

baka0815:
Sorry, I didn't get you. Yes, it *doesn't* work without any graphical backend like X11.
So X11 (or similar) is necessary.

mrfixit2001:
Well, it USED to work fine without x11... now it won't even compile without it :(

PR #1376, opened 2018-09-12
Android: Complete conversion to API 26 standards (Cont.)

AbandonedCart:


Retrogiraffee:
@AbandonedCart  Thank you for your support! 
I tried today with newest beta (r8.1-28-g6ad13268 (431475743))
The good news:
- All settings/changes are saved now.
But no matter if i chose "l/r" or "a/b" or "none" - the right analog stick is doing nothing. If i don't use the physical-settings, the right analog stick pauses the game when i click the analog stick. But moving around with right analog stick does nothing (even with or without physical settings on).

PR #1377, opened 2018-09-14
Android/Shell: File Uri and NDK changes for N+

AbandonedCart:


skmp:
LGTM. Updated title to platform/module notation for prettier change log generation.

AbandonedCart:
Android no longer supports reading files from secondary storage (ie. an sdcard in a Shield Tablet) that is not stored in the proprietary directory of the application. The workaround is to access the storage folder by the general root and allow it to automatically populate the scard folder, then unravel the content uri when sending it to the emulator native code.

AbandonedCart:
TL; DR Android has gotten stupid AF 
#👎

PR #1394, opened 2018-09-20
[OnHold] Revert "Interrupt Hack setting w/ less stigma"

AbandonedCart:
This reverts commits e5c0f0e, eebf3fc, 14fc7d9

### DIFF
--- a/core/hw/aica/aica_if.cpp
+++ b/core/hw/aica/aica_if.cpp
@@ -178,28 +178,6 @@ void aica_Term()
 
 }
 
-s32 aica_pending_dma = 0;
-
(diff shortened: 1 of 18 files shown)


AbandonedCart:
`Exception in thread "main" java.lang.RuntimeException: Could not locate the Gradle launcher JAR in Gradle distribution `
... awesome...

AbandonedCart:
Pending a replacement that doesn't require explicitly setting any options.

AbandonedCart:
So, in case anyone else runs into an issue with travis saying the gradle JAR was not found, all you have to do is go into travis and clear the cache for the specific branch and then restart the build.

PR #1401, opened 2018-09-23
Docs: Import old skmp's docs

skmp:
Source: https://web.archive.org/web/20110812025705/http://drk.emudev.org:80/blog/


skmp:
@MrPsyMan would be nice to have the ones from your blog as well

PR #1404, opened 2018-09-25
Docs: First draft of Collaboration Guidelines

skmp:
This is short version of what I had in my mind. What do you think?

Todo
- [ ] Add a section about preferring public discourse vs PMs and private chats, and explain why

### DIFF
--- a/docs/Collaboration guidelines.md
+++ b/docs/Collaboration guidelines.md
@@ -15,9 +15,9 @@ We're primarily here to make Reicast better.
 
 When discussing things in the tracker, forum or discord, try to think of the project first, and not of personal issues you might have with people.
 
-Good example: Constructive feedback is good because it makes contributors feel accepted, valued and safe. When people feel safe they are more comfortable explaining their opinion/viewpoint, and that helps make the project better
+Strive for: Thinking "Constructive feedback is good because it makes contributors feel accepted, valued and safe. When people feel safe they are more comfortable explaining their opinion/viewpoint, and that helps make the project better"
 
-Bad example: Emotionally abusing people is fun and makes my day better
+Avoid: Thinking "Emotionally abusing people is fun and makes my (the abuser's) day better"
 
 Avoid negativity, be realistic, give constructive feedback
 ---
--- a/docs/Collaboration guidelines.md
+++ b/docs/Collaboration guidelines.md
@@ -1,0 +1,51 @@
+Collaboration Guidelines
+===
+
+Be polite to each other
+---
+
+No one wants to wake up to a rude reply. Try to behave the best way you can, and show understanding when others are not. 
+
+When things escalate, try to use this document as a guideline on how to de-escalate them.
+
+Think of the project first
+---
+
+We're primarily here to make Reicast better.
+
+When discussing things in the tracker, forum or discord, try to think of the project first, and not of personal issues you might have with people.
+
+Good example: Constructive feedback is good because it makes contributors feel accepted, valued and safe. When people feel safe they are more comfortable explaining their opinion/viewpoint, and that helps make the project better
+
+Bad example: Emotionally abusing people is fun and makes my day better
+
+Avoid negativity, be realistic, give constructive feedback
+---
+
+Before replying, ask yourself the following questions.
+- How will the person I'm replying to feel after reading my comment?
+- Am I motivating them to get work done, or just spreading negativity?
+- Am I giving feedback that is valuable?
+
+Gracefully recover from missteps
+---
+
+We are all humans, and we have emotions and troubles, so often our conduct is not ideal. Here are some ideas on how to get things back on track.
+- When not thinking with a clear mind, take a break. Go for a walk, call a friend, stare at TV, smoke a cig, whatever works for you.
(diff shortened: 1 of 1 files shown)


docs/Collaboration guidelines.md
```diff
@@ -0,0 +1,51 @@
...
+- Try to convince others based on scientific/rational merit, not personal opinion or emotion. 
+- Try to extract knowledge, even when others are behaving in a not-so-nice way. Remember: almost every viewpoint has some value.
+- Try to decompose your thinking process and explain it to others, identify leaps of rationality.
+- Try to understand other's people's thinking process, and how they decompose the ideas in their heads.
+- Communicate without taking things personally
+
+A good rule of thumb is: If you can't explain why something should be done in some way, you're probably not thinking very rationally.
+
+Bad example: C++ is good because I know it
+Good example: C++ lets me do this and that, because of this and that, and I thin kit can work well because of ...
```
MrPsyMan: Typo on last line

think it can

MrPsyMan: Also, it's also a good idea to write down what we can do with people who remain negative,  aggressive,  toxic.

AbandonedCart: Isn't that how you get a member badge? 👅

AbandonedCart:
Wasn't this spawned from discussion in an issue? I can't seem to track it down.

docs/Collaboration guidelines.md
```diff
@@ -0,0 +1,51 @@
+Collaboration Guidelines
+===
+
+Be polite to each other
+---
+
+No one wants to wake up to a rude reply. Try to behave the best way you can, and show understanding when others are not.
```
AbandonedCart: This comes off a bit pessimistic. Even if the tone of this specific point can't be changed, it could be moved after the next point so it starts off positive.

docs/Collaboration guidelines.md
```diff
@@ -0,0 +1,51 @@
+Collaboration Guidelines
+===
+
+Be polite to each other
+---
+
+No one wants to wake up to a rude reply. Try to behave the best way you can, and show understanding when others are not. 
+
+When things escalate, try to use this document as a guideline on how to de-escalate them.
```
AbandonedCart: Changing this to "how to diffuse" would sound a bit clearer. Reusing the same word in a hyphenated negation tends to make it sound like there is not much clarity that can be offered about the word.

docs/Collaboration guidelines.md
```diff
@@ -0,0 +1,51 @@
...
+
+Think of the project first
```
AbandonedCart: I like this point, but think it could be a little clearer. "Put the project before your personal feelings"

docs/Collaboration guidelines.md
```diff
@@ -0,0 +1,51 @@
...
+
+Avoid negativity, be realistic, give constructive feedback
+---
+
+Before replying, ask yourself the following questions.
+- How will the person I'm replying to feel after reading my comment?
+- Am I motivating them to get work done, or just spreading negativity?
+- Am I giving feedback that is valuable?
+
+Gracefully recover from missteps
```
AbandonedCart: Another good point. You may also want to include "and don't be afraid to ask for help"

docs/Collaboration guidelines.md
```diff
@@ -0,0 +1,51 @@
...
+---
+
+We are all humans, and we have emotions and troubles, so often our conduct is not ideal. Here are some ideas on how to get things back on track.
+- When not thinking with a clear mind, take a break. Go for a walk, call a friend, stare at TV, smoke a cig, whatever works for you.
```
AbandonedCart: Might not want to include smoking. You don't want some radical pointing out that you are motivating "bad habits" (even if that is all they contribute).

docs/Collaboration guidelines.md
```diff
@@ -0,0 +1,51 @@
...
+- When in the wrong, apologize, and try to not repeat the offending behavior. Remember, an apology is sincere _only_ if it changes your behavior.
```
AbandonedCart: I think the formatting got messed up here. Were you trying to **_only_**?

docs/Collaboration guidelines.md
```diff
@@ -0,0 +1,51 @@
...
+- Defend your opinions, but be open to exploring other ideas, especially after a disagreement. "A strong opinion, weakly held"
+
+Remember, we're doing science
+---
+
+- Try to convince others based on scientific/rational merit, not personal opinion or emotion.
```
AbandonedCart: I'm not sure scientific is the best word, even though I get your meaning. It may be easier to just say "proven"

skmp: Well, scientific as in "following the scientific method". A lot of the things we discuss or plan are not proven, but they hopefully make sufficient rational sense to be reasonable.

AbandonedCart: Fair enough. That isn't a half bad way to word it, either. "Try to follow the scientific method"

baka0815:
We don't want this anymore?

PR #1406, opened 2018-09-25
AppVeyor: Enable v140 toolset

baka0815:
We already set os to VisualStudio 2015, but the toolset was still set to v120 (VS 2013).


AbandonedCart:
@dmiller423 and I made all the changes needed that there shouldn't be any issues with it publishing to the existing CI, so this shouldn't have any visible impact to the end-user. Thanks :)

PR #1409, opened 2018-09-26
Documentation Effort

luserx0:
- Description:
  * Added/Updated READMEs within multiple locations
  * Added the Covenant's Code of Conduct (for review by skmp to combine
  it with thy guidelines)
  * Added Specs about Dreamcast, Naomi
  * Updated the old blog posts imported by skmp to be a little more
  modern. Fixed some typos and renamed them to eliminate spaces

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,42 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers of the Reicast Emulator pledge to making participation in our project and our community a harassment-free experience for everyone, regardless ofage, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable behavior and are expected to take appropriate and fair corrective action in response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct, or to ban temporarily or permanently any contributor for other behaviors that they deem inappropriate, threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces when an individual is representing the project or its community. Examples of representing a project or community include using an official project e-mail address, posting via an official social media account, or acting as an appointed representative at an online or offline event. Representation of a project may be further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by contacting the project team at skmp@emudev.org and lx0@emudev.org. The project team will review and investigate all complaints, and will respond in a way that it deems appropriate to the circumstances. The project team is obligated to maintain confidentiality with regard to the reporter of an incident. Further details of specific enforcement policies may be posted separately.
+
+Project maintainers who do not follow or enforce the Code of Conduct in good faith may face temporary or permanent repercussions as determined by other members of the project's leadership.
+
+**Cheerios**,
+The Reicast Dev Team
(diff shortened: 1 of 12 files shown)


docs/DreamCast_Specs.md
```diff
@@ -0,0 +1,22 @@
+# SEGA Dreamcast Specifications
```
skmp: Why not name this "Dreamcast_Overview.md" to match the naomi one?

skmp:
Not sure the `_`s look better than ` `. What is the rationale behind them?

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,42 @@
+# Contributor Covenant Code of Conduct
```
skmp: This looks fairly interesting. I think we should merge this with #1404.

core/README.md
```diff
@@ -1,4 +1,17 @@
-core
+# core/libdreamcast
 ===========
 
-All of the interesting bits are here
\ No newline at end of file
+Here lies the core of our codebase. Everything that's OS inspecific rests here.
```
skmp: Wouldn't this better be handled via doxygen or such? Looks a bit too detailed to be in the readme.

core/emitter/README.md
```diff
@@ -0,0 +1,6 @@
+# Our Emitter
```
skmp: Same comment regarding doxygen

core/hw/gdrom/GD-ROM_Overview.md
```diff
@@ -0,0 +1,8 @@
+# GD-ROM: Gigabyte Disc Read-Only Memory
```
skmp: Same comment regarding doxygen

CODE_OF_CONDUCT.md
AbandonedCart: Contributor Covenant Code of Conduct seems like a mouthful. Couldn't it be shortened to Contributor Code of Conduct? I always think of pentagrams and black cats when I hear covenant.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers of the Reicast Emulator pledge to making participation in our project and our community a harassment-free experience for everyone, regardless ofage, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
```
AbandonedCart: Spotted a typo.
"regardless of age"

luserx0: Yup saw it too. Fixed ^_^

core/README.md
luserx0: Well, we do need to have a manual draft of what the automated documentation tool should produce us

docs/DreamCast_Specs.md
luserx0: Because in this doc there are only the Machine's specs, in Naomi there's a bit more info.

CODE_OF_CONDUCT.md
luserx0: Agreed. On both.

AbandonedCart: After a lot of thought, I don't think merging the two together would keep the same purpose of each as individual documents. In a business setting, this would be the Code of Conduct, but the other one would be the ethics and culture statements. Both attempt to lead people in the right direction, but have two completely different tones. Rather than trying to weave them together, I think putting the main points of the other document as a subsection about project etiquette would maintain the general feel of both without losing the intention of either.

luserx0: I have to slightly disagree on this. 

In a FOSS project you need both extensive but not tiring documentation. 

A Code of Conduct achieves exactly and - just - that: introducing people to the ethics of a community. In our case, the reicast community. Establishing the proper way of communicating and collaborating means also understanding the ethics and culture of a community. 

Contributor Code of Conduct is describing the doc on point.

AbandonedCart: @skmp recommended merging the two. I am stating that they are two completely different tones and the "new and important" points of the other one may be better suited as a subsection of this, more professionally worded, document over flat out combining them.
You may also want to take into consideration the fact that this document seems more geared at contributors / members, while the other covers a broader range of communications.

core/imgread/common.cpp
```diff
@@ -131,7 +131,7 @@ bool ConvertSector(u8* in_buff , u8* out_buff , int from , int to,int sector)
 
 Disc* OpenDisc(const wchar* fn)
 {
-	Disc* rv = nullptr;
+	Disc* rv = 0;
```
baka0815: This should be fixed in a separate PR, not in a "documentation fixes" PR.

luserx0: The two PRs' commits have been mixed up.
This is originally included in the Snapcraft fixes PR.
Let me fix this real quick and review again.

baka0815: Yes, please move this commit to #1430 as that would be where it belongs.

baka0815: It looks loke you're no longer working on this? There are more issues regarding this single problem.
I might reopen my fix to address this single problem.

baka0815:
I think you should remove all commits after the initial one in this PR. The fix for the ``nullptr`` is no longer necessary and the 'testing' and 'still testing' commits look like they don't belong into this PR I guess.

core/imgread/common.cpp
baka0815: This is no longer necessary as it's fixed in master.

PR #1412, opened 2018-09-26
Schedule the interrupt signaling the end of an Aica DMA transfer

flyinghead:
Schedule the interrupt signaling the end of an Aica DMA transfer instead of raising it immediately.

Fixes Street Fighter Alpha 3 and Bomberman Online

### DIFF
--- a/core/nullDC.cpp
+++ b/core/nullDC.cpp
@@ -297,6 +297,7 @@ void LoadSettings()
 	settings.aica.LimitFPS			= cfgLoadInt("config", "aica.LimitFPS", 1);
 	settings.aica.NoBatch			= cfgLoadInt("config", "aica.NoBatch", 0);
 	settings.aica.NoSound			= cfgLoadInt("config", "aica.NoSound", 0);
+	settings.aica.OldSyncronousDma	= cfgLoadBool("config", "aica.OldSyncronousDma", false);
 	settings.rend.UseMipmaps		= cfgLoadInt("config", "rend.UseMipmaps", 1);
 	settings.rend.WideScreen		= cfgLoadInt("config", "rend.WideScreen", 0);
 	settings.rend.Clipping			= cfgLoadInt("config", "rend.Clipping", 1);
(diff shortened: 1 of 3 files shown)


skmp:
This is definitely a step in the right direction. There's various nuances that we should implement, or at least assert about, like dma pauses, g1/g2 locks, and dma aborts. 

This is much cleaner than other diffs I've seen trying to make DMAs take time.

@flyinghead do you want to add detection for pauses / aborts before merging this in? I'm fine with us handling it at a later time. Created a general follow up ticket, #1414.

@MrPsyMan do you think this should be included in the 18.10 release? I'd be wary to push it out without a thorough pass of 60/70 games myself, because of how easily timing can break games, and how much unimplemented functionality is exposed by making the DMA not be instant. Am I too cautious?

AbandonedCart:
> how about adding an option for it, "AsyncAICADma", and enable it with some per-game configs? Then even if it causes issues users will have a workaround, and they can turn it if they want to test for other games.

Isn't the purpose of this to do away with needing to configure it at all and actually **fix** the functionality, though? It seems like that would be 2 steps forward, one step back.

MrPsyMan:
I'd say to add it as an option enabled by default, until more related functionality is implemented. 

If it turns out it breaks something, it'd be far easier to pinpoint if this is the cause when the users can turn it off.

core/nullDC.cpp
```diff
@@ -297,6 +297,7 @@ void LoadSettings()
 	settings.aica.LimitFPS			= cfgLoadInt("config", "aica.LimitFPS", 1);
 	settings.aica.NoBatch			= cfgLoadInt("config", "aica.NoBatch", 0);
 	settings.aica.NoSound			= cfgLoadInt("config", "aica.NoSound", 0);
+	settings.aica.OldSyncronousDma = cfgLoadBool("config", "aica.OldSyncronousDma", false);
```
baka0815: Indentation.

As much as I dislike the tabular layout it currently is, it, well, currently is and should therefor stay as it is or be reformatted as a whole.

baka0815: Oh well, I just switched the diff to a bigger viewpoint and saw that there already are many different identations here...

PR #1413, opened 2018-09-27
Only default to a storage list when not searching

AbandonedCart:
This one missed the bus...


AbandonedCart:
... looks like I need to open another PR...

PR #1418, opened 2018-09-27
Convert to compat tints instead of style definitions

AbandonedCart:


AbandonedCart:
Lock in your theme ideas now. This is going to be a pita to maintain.

PR #1419, opened 2018-09-27
Theme improvements and updates

AbandonedCart:
The wood grain effect ended up looking like awkward tiles. It should be cleaner


AbandonedCart:
screen shot 2018-09-27 at 2 32 00 pm

More like waves, less like blobs

baka0815:
The new one is way clearer and not so disruptive.

PR #1422, opened 2018-09-30
Avoid duplicate URI references, Verify and verify again

AbandonedCart:


shell/android-studio/reicast/src/main/java/com/reicast/emulator/emu/GL2JNIView.java
```diff
@@ -659,7 +659,7 @@ void Die() {
         }
 
         void reiosInfo(String reiosId, String reiosSoftware) {
-            if (fileName != null) {
+            if (fileName != null && !reiosId.equals("") && !reiosSoftware.equals("")) {
```
baka0815: Something I remember from my time writing Java code is that it's a good idea to reverse the equals check for an empty string to remove possible nullpointer exceptions.

``"".equals(reiosId)``

AbandonedCart: `fileName` is the only value that initializes with a `null`

AbandonedCart: In all reality, though, I was trying to work out the logic for how to avoid even getting here with a blank. No reason to get all the way to the shell just to throw it out.

PR #1423, opened 2018-09-30
Add cable/region/broadcast to per-game configuration.

flyinghead:
Do not save per-game settings if the game name or id is unknown (NAOMI).


AbandonedCart:
Thanks for this and for getting it up so fast. I followed up with the Android UI. If you need it, the specific commit is https://github.com/reicast/reicast-emulator/commit/dc5a3df78fbdfd30a64f9b7d5172fc6fcd54e9d8

Issue #1410, opened 2018-09-26
Sonic Adventure 2 freezing after Green Forest


vladone97:
**Please Note: This form is the minimum required information for submitting bugs.**  
**Removing this form may lead to your issue being closed until it is completed.**

| Platform | Branch | Hash | CIDL |
| ------ | ------ | ---- | ---- |
| Android 7.0 | Google Play | Hashtag (skip if unknown) | Downloaded from build site? |

**_Description of the Issue_**

Sonic Adventure 2 froze on the Hero story, right after I finished Green Forest. It just shows a black screen forever. The same thing happens if I try to select the scene that comes right after Green Forest on a completed save file.

**_Debugging Steps Tested_**

  * reloaded the save file and tried to continue where I left off in the Hero story
  * selected the scene on a completed save file
  * tried to load the scene with PVR rendering disabled

**_Logs Gathered_**

```

Please paste the contents of the log / logs here
You may leave this field blank if you have none.

```

**_Screenshots_**

(Replace this line with any issue screenshots)

- [X ] Place an X in the box to confirm you have filled in this form

MrPsyMan:
Are you running a gdi or cdi image of the game?

vladone97:
But I used the same image on older versions of Reicast and it was working back then.

MrPsyMan:
Then we want the testing team to confirm this. @baka0815 any thoughts on this?

flyinghead:
> Team: I consider adding a warning whenever users load non gdi images.
> Opinions on the matter?

There is one already https://github.com/reicast/reicast-emulator/commit/1a4f597bd814915f1bd6f188c575cdfead4c73bc
Personally I dislike the idea of having a pop up every time I load a CDI. Some are bad but most CDIs are fine.

MrPsyMan:
I can't say I like the idea, but it's better than keeping users in the dark.

AbandonedCart:
In Android, the popup is a logcat warning. It's there if you look, but non-invasive.

MrPsyMan:
So users are left in the dark...

I'd rather annoy users a bit to help them, rather than having a bad rip ruin their gaming session. 

That's just my opinion though.

MrPsyMan:
Homebrews and the likes are in cd images so we don't really have a choice.

baka0815:
What about:
[...]

PR #1424, opened 2018-10-01
Null checks, Prettier user-end appearance, etc

AbandonedCart:


AbandonedCart:
This avoids changing any values until it can be verified how they were originally determined. It's been 4 years since they were originally implemented and the change was picked from a remote that is no longer stored on Github. For now, given the limited usage, they are being left as-is.

AbandonedCart:
The above change will have a single use today, but the base concept can also be leveraged for a lot of other things in the future.

shell/android-studio/reicast/src/main/java/com/reicast/emulator/config/OptionsFragment.java
```diff
@@ -767,4 +764,34 @@ private void showToastMessage(String message, int duration) {
 				.getDimensionPixelOffset(R.dimen.snackbar_icon_padding));
 		snackbar.show();
 	}
+
+	private int getBroadcastValue(String broadcastName) {
+		if (broadcastName.equals("NTSC-J"))
```
baka0815: Isn't that something that should be done inside reicast core to avoi duplicate code?

shell/android-studio/reicast/src/main/java/com/reicast/emulator/Emulator.java
```diff
@@ -107,6 +107,28 @@ public void loadGameConfiguration(String gameId) {
         JNIdc.bootdisk(mPrefs.getString(pref_bootdisk, bootdisk));
     }
 
+    public void isVGACompatible(String gameId) {
```
baka0815: Same as my other comment.

shell/android-studio/reicast/src/main/java/com/reicast/emulator/config/OptionsFragment.java
AbandonedCart: It's only used in two places and under review. If the values are incorrect, the replacement code does not require any sort of helper. It was done this way to make for a simple removal.

shell/android-studio/reicast/src/main/java/com/reicast/emulator/Emulator.java
AbandonedCart: This is as general as it gets. This class extends Application.

AbandonedCart:
Important note: This only includes NA / EU releases. The forum hints to a list of NTSC-J games, but it was not linked and will take some digging to determine if it was ever made. Ideally `-1` will just default to the user setting.

AbandonedCart:
Switched the values to favor patchable over TV-only, since that is an "upgrade"

AbandonedCart:
This prevents the user from being forced into a setting, but still guides them to a default based on what testing has proven is compatible.

AbandonedCart:
```
                PAL-E - Conflict Zone - GOT - [DCCM]
                NTSC-U - Donald Duck: Goin' Quackers - GOT - [DCCM/RDC]
                NTSC-U - ESPN International Track n Field - GOT -[DCCM]
                PAL-E - Millenium Soldier: Expendable (Expendable) - GOT - [DCP]
                PAL-E - Heavy Metal: Geomatrix - GOT - [DCP]
                NTSC-U - House of the Dead 2 - GOT - [DCCM]
                NTSC-U - Loony Toons Space Race - GOT - [DCRES/OVERRiDE]
                PAL-E - Midway Greatest Arcade Hits Volume 1 - GOT - [DCCM]
                PAL-E - Mortal Kombat Gold - GOT - [DCP]
                PAL-E - NBA Hoopz - GOT - [PULSAR]
                NTSC-U - NHL 2K - GOT - [DCCM]
                PAL-E Freestyle Scooter [Razor Freestyle Scooter]
                NTSC-U Record of Lodoss War
                NTSC-U Roadsters
                PAL-E San Fransisco Rush 2049
                NTSC-U Sega Bass Fishing
                NTSC-U Sega Rally Championship 2
                NTSC-U Silent Scope
                PAL-E Speed Devils Online Racing
                PAL-E Star Wars: Demolition
                PAL-E Star Wars: Episode 1 Jedi Power Battles
                PAL-E Star Wars: Episode 1 Racer
                PAL-E Street Fighter III: 3rd Strike
                PAL-E Suzuki Alstare Extreme Racing
                PAL-E Tech Romancer
                PAL-E Tokyo Highway Challenge 2 [Tokyo Extreme Racer 2]
                NTSC-U Toy Comander
                NTSC-U Toy Story 2: Buzz Lightyear to the Rescue!
                PAL-E TrickStyle
                PAL-E Urban Chaos
                NTSC-U Vigilante 8: 2nd Offense
                NTSC-U Wetrix+
                NTSC-U Xtreme Sports
                NTSC-U Zombie Revenge
```

This is the list of games (from the comments) that were never tested. IDs will be added once the current progress has been verified.

AbandonedCart:
For simplicity, this now returns 3 (TV / emu default) for unknown. No use returning -1 just to return an alternate value, only to read it as 3 anyway.

AbandonedCart:
@flyinghead I believe you use a hybrid version of the per-game settings that sets the default based on the game, but then allows overriding it with a configured value. I think this should match Android up to much what you were already doing with the addition of the VGA hoopla.
I know you mentioned there being some games that required certain settings, so hopefully the testing done for the list had the same results.

baka0815:
> This is the list of games (from the comments) that were never tested.
> They are currently set to default, but could be updated in the future.

What would be necessary to confirm they work?

AbandonedCart:
Set the cable to VGA and see if the graphics still work.

AbandonedCart:
Adding unconfirmed items to a final list and converting this for other platforms can be a follow-up. This is already getting a bit lengthy. Assuming the toast are not displaying because they had a limited context, this should also address the discussion of how to notify the user about CDI in https://github.com/reicast/reicast-emulator/issues/1410

flyinghead:
@AbandonedCart Yes, I originally backported libretro per-game settings, which are hard-coded and non overridable I believe. With the PGC in emu.cfg and in the android UI, there is probably redundant stuff. More technical debt...

AbandonedCart:
I think it's actually more effective that way. Set a values based on what we know, but let the user change it if they disagree (or some other factor comes into play)

PR #1425, opened 2018-10-02
Attempt to update Dropbox API from 1.6 to 3.0

AbandonedCart:
Not entirely sure if there is more to the login. Their documentation is a Javadoc and the samples are written like test modules meant to prove it functions with the least amount of code.


AbandonedCart:
`The number of method references in a .dex file cannot exceed 64K.`
Between 20K - 30K are just the Dropbox SDK alone.

AbandonedCart:
As a bonus, the compiled APK is 1.6 MB smaller overall **after** the addition of the absurdly bloated Dropbox SDK.

shell/android-studio/reicast/src/main/java/com/reicast/emulator/FileBrowser.java
```diff
@@ -194,6 +194,11 @@ private void installButtons() {
 			if (buttons != null && buttons.exists()) {
 				in = new FileInputStream(buttons);
 			} else if (!file.exists() || file.length() == 0) {
+				try {
+					file.createNewFile();
+				} catch (Exception e) {
+					// N+ files be broken
```
skmp: Files are broken?

AbandonedCart:
Dropbox upload / download is now working. Just in time for the hurricane...

PR #1434, opened 2018-10-15
Don't initialize the disc with a nullptr as that's not always available

### LINKED ISSUE #1375: Fix the snapcraft build
snapcraft build is broken because of c++11 usage

Just found out today, we haven't been pushing snapcraft builds since Aug21. 

We need to figure out how to fix this

build log
https://build.snapcraft.io/user/reicast/reicast-emulator/331060

:(

baka0815:
Fixes #1375


luserx0:
This is fixed in my PRs, awaiting your (meaningful) reviewal since last week.
Closing this down as duplicate.

Issue #1229, opened 2018-07-21
Reicast fails to render some KallistiOS scenes


Protofall:
Some of my KallistiOS programs, that I have tested on real Dreamcast hardware, fail to render anything in the reicast emulator. Attached is an example CDI of a program that does this. The functions that load the data into memory all return 0 (Success). The problem could have to do with rendering. Also note every program I tested that did work uses PNG assets that are converted into a DC usable format at runtime whereas this program uses the DTEX and DTEX.PAL formats which are loaded straight into memory without needing to convert them.

In this program I draw a 256 by 256 PAL4BPP DTEX spritesheet in the top left corner. Attached is a screenshot of the lxdream emulator running the same program along with a download link for the CDI file

![proglxdream](https://user-images.githubusercontent.com/20989925/43035832-0db4dc0a-8d35-11e8-9d11-37ddd91de0b4.png)

http://www.mediafire.com/file/hm3724c2m7hutyc/Prog.cdi/file

baka0815:
Just to be sure, the expected output is the one from the screenshot, right?

Protofall:
That is correct.

Protofall:
I've ran both the master and @DevilFrost 's fork with multiple KOS tests (Including Prog.cdi and KOS' packed in PNG example that has an opaque background and custom text on translucent polys). Although the main branch does mess up a few times (Such as failing to draw the PNG example's background), the fork messes up the same things and sometimes more (in PNG it draws nothing). The only time I'm seeing Devil's fork out performing master is with Prog.cdi (And other programs using the draw function from Prog.cdi).

If it helps this is the draw function used in Prog.cdi and programs that use Prog.cdi's draw function

`uint8_t graphics_draw_sprite(const struct crayon_details *details, float draw_x, float draw_y, float draw_z,
    float scale_x, float scale_y, uint16_t frame_x, uint16_t frame_y, uint8_t paletteNumber){
 
    //Screen coords. letter0 is top left coord, letter1 is bottom right coord. Z is depth (Layer)
    const float x0 = draw_x;
    const float y0 = draw_y;
    const float x1 = draw_x + (details->width) * scale_x;
    const float y1 = draw_y + (details->height) * scale_y;
    const float z = draw_z;
 
    //Texture coords. letter0 and letter1 have same logic as before
    const float u0 = frame_x / (float)details->texture_dims;
    const float v0 = frame_y / (float)details->texture_dims;
[...]

PR #1440, opened 2018-10-24
Fix reicast fails to render some KallistiOS scenes

DevilFrost:


skmp:
@DevilFrost what is the rationale behind this? Why the change applies only on the FRONT face culling? As this affects behavior that most games use, have you tested it?

DevilFrost:
@skmp this is a fix for Issue [#1229](https://github.com/reicast/reicast-emulator/issues/1229)

flyinghead:
https://github.com/reicast/reicast-emulator/commit/6ffe4c0280ef483527cf804f44f8960ecfef7aa0#diff-566724dab6ce076b0b11d297ea1167b6 might be a better fix

DevilFrost:
@skmp need to test games.

dmiller423:
@DevilFrost You would need to do some testing,  not skmp if that's what you meant.  As well as some explanation as to why this is only done for translucent polys w. autosort on...  Perhaps the games flyinghead said just the xor already fixes are a good start:  "Fixes South Park Rally and Wacky Races invisible menus."

flyinghead:
The commit I'm mentioning fixes a problem that affects all Sprites: their face orientation is reversed. This can be seen by checking the order of vertices in the resulting poly. Rather than putting the vertices in the right order, I just reversed the cull mode with the same result.

DevilFrost:
@dmiller423 Prog.cdi draws a translucent polys autosort on and GL_FRONT. With GL_FRONT incorrect. Replacing an GL_FRONT with an GL_BACK solves the problem. I tested the Wacky Races. Menu in the Wacky Races is drawn correctly. With the fix of flyinghead, everything is also drawn correctly.

I think fix flyinghead is better.

Protofall:
I've run some KOS tests with @DevilFrost 's fork and most of the time I'm finding master is more accurate. I made a more detailed analysis in the comments of Issue #1229 (Posted there because it gives better insight as to why master can't render Prog.cdi)

dmiller423:
I'm not surprised,  the xor FH uses for sprites only is a bit of a hack but is more plausible without knowing what else might use similar behaviour...   It's been a long time since I've looked at this area, and I was quite busy in those days trying to fix Z-issues and implement mod volumes and such.   This is why I said it has to be tested fully,  thanks for the input @Protofall

PR #1448, opened 2018-10-28
Enable log2 depth at runtime when supported (GL, GLES 3)

flyinghead:
Fix issues with pixel scaling and scissoring.
Wacky Races should now be playable.


dmiller423:
Good work 👍

PR #1452, opened 2018-10-30
Snap Additions

luserx0:
Description: 
- Added README for snap, 
- played around with snapcraft.yaml 
- and updated root README with snap info

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,18 @@
+[![Snap Status](https://build.snapcraft.io/badge/reicast/reicast-emulator.svg)](https://build.snapcraft.io/user/reicast/reicast-emulator)
+
 reicast
 ===========
-reicast is a multi-platform Sega Dreamcast emulator.
+**reicast** is a multi-platform Sega Dreamcast emulator.
 
 This is a developer-oriented resource, if you just want bins head over to http://reicast.com/
 
 For development discussion, join [#reicast in freenode](https://webchat.freenode.net/?channels=reicast)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,19 +1,18 @@
...
 ===========
-reicast is a multi-platform Sega Dreamcast emulator.
+**reicast** is a multi-platform Sega Dreamcast emulator.
 
 This is a developer-oriented resource, if you just want bins head over to http://reicast.com/
 
 For development discussion, join [#reicast in freenode](https://webchat.freenode.net/?channels=reicast)
  or stop by the [reicast Discord server](http://discord.gg/Hc6CF72)
 
-Caution
```
baka0815: Why are you removing this?

snap/snapcraft.yaml
```diff
@@ -1,12 +1,13 @@
 name: reicast
 version: git
-summary: Reicast
+summary: Reicast -- The fastest Dreamcast Emulator
```
baka0815: Do we really want to state that?

luserx0: Yup ;) We are and we take pride in it

luserx0: It's also only stated in the snapcraft yaml, so nonarcissistic harm done :)

README.md
luserx0: There's no need to add such a caution these days. We are making things more documented and stable. No need to scare off ppl

PR #1460, opened 2018-11-11
Multipass render to texture with options - implemented

mar753:
I was a huge fan of THPS2, so many memories ;), but when I saw those artifacts on my Android phone:
screenshot_2018-11-11-21-41-36-881_com reicast emulator
I decided that it must be fixed!
Current version of reicast supports GLES 2.0. Plain glReadPixels() operation is time consuming, sometimes preventing smooth 60fps gameplay. I have even upgraded GLES to 3.0 in reicast application and tried using single PBO (on GLES 2.0 we can use eglCreateImageKHR() as well, but unfortunately GraphicBuffer is not available in NDK, without compiling with Android native code, fortunately solution is to dynamically load libui.so library and get GraphicBuffer from there, check this out: https://github.com/fuyufjh/GraphicBuffer), but performance increase was not noticeable, at least on my device with Qualcomm Snapdragon 625.

I know that 'render to texture' support was disabled because of performance issues, so I have prepared several options, what I hope will be a good solution. There is a new menu select (Experimental section) called 'Render to texture', where available options are:
- 'Disabled - skip frames' - this option skipes rtt frames, exactly as it is working now (image above)
- 'Zeros' - this option writes zeros to the texture (for all RGBA channels) that will be used as shadow/anything else
screenshot_2018-11-11-21-39-49-182_com reicast emulator
- 'Ones' -  this option writes ones to the texture (for all RGBA channels) that will be used as shadow/anything else
screenshot_2018-11-11-21-43-12-981_com reicast emulator
- 'Shadow circle' - (default) this option creates a circle (disk in fact) in a 128x128 buffer using sin/cos functions and writes this data to the texture, which is then used as a nice pseudo-shadow
screenshot_2018-11-11-21-44-55-578_com reicast emulator
- 'Full' - this option implements full 'render to texture' support with two pass rendering (first: rtt, second: scene)
screenshot_2018-11-11-21-46-26-977_com reicast emulator

In terms of performance: 
when 'render to texture' is 'Disabled - skip frames' performance is exactly the same as it is now in the current version (it is using exactly the same code as currently), there is also no noticeable performance differencies in every other modes/options, except 'Full' (Drawing is disabled in every mode except 'Full'), which is sometimes much slower. Maybe, limiting the draw distance may help here.
When using 'Full', only two pass rendering is available at the moment. It will be nice to implement multipass RTT as well.

Feel free to fetch this branch and test!

Cheers,
Marcel


skmp:
This pull request **introduces 2 alerts** when merging 4d39c10c9151e1b75697240c98aad03fcd01314a into 19ca0edc2f005f83be39b042164e064f4bb3b877 - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-048a359cca7312b723742326d657a9999c7d05ee)

**new alerts:**

* 1 for Multiplication result converted to larger type
* 1 for Short global name

---

mar753:
Thank you for your feedback @flyinghead!

Regarding my implementation, I have done some debugging before and yes, I know that byte swap from red channel to alpha without LSB (RGB565 -> 'ARGB4444') needed by THPS2 can be done automatically when we read pixels directly to vram[fb_rtt.TexAddr<<3] location without any swapping, but it must be done within the same RTT frame rendering cycle (like in your code). I have even determined that fb_rtt.TexAddr<<3 for the shadow in THPS2 equals 0x4D0180, at least on my Android device, this can be used for debugging purposes.

In my implementation I have moved glReadPixels() to the next frame rendering cycle on purpose (after RTT drawing cycle), doing like this I was able to significantly increase the performance, thus framerate (tested on Qualcomm Snapdragon 625) with fully enabled single pass RTT (first: RTT frame, next: scene frame, then: RTT frame etc.). Performance is better because during making some stuff on CPU to prepare the next frame, RTT rendering/drawing from the previous frame is executing in parallel. In contrast, calling glReadPixels() within an RTT frame rendering cycle, causes a stall on glReadPixels() until rendering/drawing is finished. 

Unfortunately this approach forces me to make this color channel swap manually (red->alpha), because when a next frame is being processed (scene drawing), vram[fb_rtt.TexAddr<<3] location is not valid anymore - now it is valid for a current frame thus I cannot write to it. But... I am doing:

	for (u32 i = 0; i < w * h; i++)
	{
		buf[i] = ((*dataPointer & 0xF000) >> 12) | ((*dataPointer & 0x0FFF) << 4);
		*dataPointer++;
	}
	glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, w, h, 0, GL_RGBA, GL_UNSIGNED_SHORT_4_4_4_4, buf);

it is exactly what will be done in `TextureCacheData::Update()` from gltex.cpp:

	if (texID) {
		//upload to OpenGL !
		glBindTexture(GL_TEXTURE_2D, texID);
		GLuint comps=textype==GL_UNSIGNED_SHORT_5_6_5?GL_RGB:GL_RGBA;
		glTexImage2D(GL_TEXTURE_2D, 0,comps , w, h, 0, comps, textype, temp_tex_buffer);
		if (tcw.MipMapped && settings.rend.UseMipmaps)
			glGenerateMipmap(GL_TEXTURE_2D);
	}

little bit higher in this function there is a code responsible for red->alpha channel swap:

	if(texconv!=0)
	{
		texconv(&pbt,(u8*)&vram[sa],stride,h);
	}

What is more `gl_GetTexture()` at the beginning has:
	
	if (tcw.TexAddr==fb_rtt.TexAddr && fb_rtt.tex)
	{
		return fb_rtt.tex;
	}

and will always return here in my solution (when using RTT), because `fb_rtt.tex` will always be defined, so `TextureCacheData::Update()` will not be executed each time, what will be unnecessary time waste.

What is more I have limited glDetele* commands in `BindRTT()`, performance increase is rather not nociceable but maybe will be in some specific conditions.

But you are right in terms of hardcoding red->alpha swap/move in my code, because when a texture to which we are rendering is RGB565 I am always doing the swap and converting the texture to RGBA4444, what can cause artifacts. I have tested several games and the problem was not visible, probably because they were using some other formats for RTT rendering like RGBA5551 or RGBA4444 from the beginning. Anyway, the artifacts can be visible only when RTT is in use, thus changing 'Render to texture' select to 'Zeros' in my implementation will prevent any artifacts, because in case of RGB565, rendered texture will be transparent, so it will not be visible anyways, if I am thinking correctly. I can make this option a default one.

Do you know maybe games that uses RGB565 for RTT rendering in the 'normal' way that I can test (maybe some games with mirrors)?

flyinghead:
Crazy Taxi uses RTT for the pause menu in game.
Many games use RTT for crossfade transitions. See the opening demos of Skies of Arcadia and Virtua Tennis for example.
Rez makes heavy use of RTT for graphic effects and level transitions.

mar753:
Render to texture (RTT) with multipass rendering support has been implemented. 
During processing RTT frame, drawing is initiated, but glReadPixels() is invoked in the next nearest non-RTT frame - where RTT texture is needed for screen rendering, in this way GPU and CPU can work simultaneously and stalls are minimized. This improves the performance significantly. It is recommended to enable 'Synchronous rendering' as well - I saw some issues with level transitions in Rez without this option enabled.

OpenGL ES 2.0 fully compatible solution was implemented, when reading pixels from a framebuffer object (FBO).
For example, when a texture (or color renderbuffer) bound to the FBO is RGBA5551 format/type, GL_IMPLEMENTATION_COLOR_READ_TYPE query is executed just before glReadPixels() to explore GPU's supported image type, when GL_UNSIGNED_SHORT_5_5_5_1 will be returned, we proceed the standard way, but if something other is returned GL_UNSIGNED_BYTE type and GL_RGBA format are used to read pixels - OpenGL ES 2.0 compatible hardware must always support this format/type. We need to properly convert it back to 16 bit color value then. This fixes an issue with ARM Mali-450 MP GPU, where (as I saw) GL_RGB/GL_UNSIGNED_SHORT_5_6_5 is the only 16 bit format/type supported to read back from a framebuffer object.

There is an issue on Adreno 506 with RGBA5551 format/type, to be specific I was able to observe a huge performance drop when using this image format/type, even though this GPU declares that it is supported (GL_IMPLEMENTATION_COLOR_READ_TYPE query). This format/type is supported, but probably there is some time expensive internal format conversion, which makes it unusable, at least here. We can see this issue very clearly in case of Rez game. Using framebuffer object with GL_UNSIGNED_BYTE texture, and then reading it using glReadPixels() with the same image format/type combination seems to be a workaround for this issue. It is strange that 32 bit color rendering works faster than RGBA5551 16 bit rendering, but it is true in this case. I have not altered TextureCacheData struct (gltex.cpp) and TexCache.h/cpp files to adapt them to support 32 bit colors, because I decided it will be a lot of changes just for this fix and because we do not need 'texconv' in this case, 32 bit image is directly written to the texture after read (bypassing TextureCacheData's 'Update()' method).

To achieve this blur effect in Rez:

screenshot_2018-12-28-14-28-11-662_com reicast emulator

blending feature is used in gldraw.cpp:
`glBlendFunc(SrcBlendGL[gp->tsp.SrcInstr],DstBlendGL[gp->tsp.DstInstr]);`
with both parameters set to GL_ONE (additive blending).
For this to work properly with different formats/types of images I needed to modify BindRTT() function code to create texture/renderbuffer/framebuffer and initialize it only once at the initialization state of a given texture address.

I have added RTT swizzle texture support too. It is used in Virtua Tennis during logo and demo video transitions to achieve 640 pixels in width of a texture.

As I saw in the source code, there is no standard code convention and a lot of a C++ code is mixed with a C code, but I have tried to keep the similar convention and limit the refactoring.

Default option for 'Render to Texture' is set to 'Zeros'. 'Full' option must be selected to enable multipass RTT.

Solution was tested with several games that use RTT (including: Rez, Virtua Tennis, THPS2 and Crazy Taxi) and is working fine.

Testing platform:
- Huawei P8 lite, Android 6 (ARM Mali-450 MP)
- Xiaomi Redmi Note 4 Snapdragon, Android 7 (Adreno 506)
- Xiaomi Pocophone F1, Android 9 (Adreno 630)

Fell free to test.

mar753:
Render to texture (RTT) with multipass rendering support has been implemented. 
During processing RTT frame, drawing is initiated, but glReadPixels() is invoked in the next nearest non-RTT frame - where RTT texture is needed for screen rendering, in this way GPU and CPU can work simultaneously and stalls are minimized. This improves the performance significantly. It is recommended to enable 'Synchronous rendering' as well - I saw some issues with level transitions in Rez without this option enabled.

OpenGL ES 2.0 fully compatible solution was implemented, when reading pixels from a framebuffer object (FBO).
For example, when a texture (or color renderbuffer) bound to the FBO is RGBA5551 format/type, GL_IMPLEMENTATION_COLOR_READ_TYPE query is executed just before glReadPixels() to explore GPU's supported image type, when GL_UNSIGNED_SHORT_5_5_5_1 will be returned, we proceed the standard way, but if something other is returned GL_UNSIGNED_BYTE type and GL_RGBA format are used to read pixels - OpenGL ES 2.0 compatible hardware must always support this format/type. We need to properly convert it back to 16 bit color value then. This fixes an issue with ARM Mali-450 MP GPU, where (as I saw) GL_RGB/GL_UNSIGNED_SHORT_5_6_5 is the only 16 bit format/type supported to read back from a framebuffer object.

There is an issue on Adreno 506 with RGBA5551 format/type, to be specific I was able to observe a huge performance drop when using this image format/type, even though this GPU declares that it is supported (GL_IMPLEMENTATION_COLOR_READ_TYPE query). This format/type is supported, but probably there is some time expensive internal format conversion, which makes it unusable, at least here. We can see this issue very clearly in case of Rez game. Using framebuffer object with GL_UNSIGNED_BYTE texture, and then reading it using glReadPixels() with the same image format/type combination seems to be a workaround for this issue. It is strange that 32 bit color rendering works faster than RGBA5551 16 bit rendering, but it is true in this case. I have not altered TextureCacheData struct (gltex.cpp) and TexCache.h/cpp files to adapt them to support 32 bit colors, because I decided it will be a lot of changes just for this fix and because we do not need 'texconv' in this case, 32 bit image is directly written to the texture after read (bypassing TextureCacheData's 'Update()' method).

To achieve this blur effect in Rez:

screenshot_2018-12-28-14-28-11-662_com reicast emulator

blending feature is used in gldraw.cpp:
`glBlendFunc(SrcBlendGL[gp->tsp.SrcInstr],DstBlendGL[gp->tsp.DstInstr]);`
with both parameters set to GL_ONE (additive blending).
For this to work properly with different formats/types of images I needed to modify BindRTT() function code to create texture/renderbuffer/framebuffer and initialize it only once at the initialization state of a given texture address.

I have added RTT swizzle texture support too. It is used in Virtua Tennis during logo and demo video transitions to achieve 640 pixels in width of a texture.

As I saw in the source code, there is no standard code convention and a lot of a C++ code is mixed with a C code, but I have tried to keep the similar convention and limit the refactoring.

Default option for 'Render to Texture' is set to 'Zeros'. 'Full' option must be selected to enable multipass RTT.

Solution was tested with several games that use RTT (including: Rez, Virtua Tennis, THPS2 and Crazy Taxi) and is working fine.

Testing platform:
- Huawei P8 lite, Android 6 (ARM Mali-450 MP)
- Xiaomi Redmi Note 4 Snapdragon, Android 7 (Adreno 506)
- Xiaomi Pocophone F1, Android 9 (Adreno 630)

Feel free to test.

skmp:
This pull request **introduces 3 alerts** when merging 302bfcd58273d37c9809b365fec87b56e88a05eb into bbc54e4125745413b4fe60b936ace1a1bf28fa31 - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-0a2207023691e2e2b77a9affc55e962cd40888be)

**new alerts:**

* 2 for Local variable hides global variable
* 1 for Multiplication result converted to larger type

---

mar753:
Render to texture (RTT) with multipass rendering support has been implemented. 
During processing RTT frame, drawing is initiated, but glReadPixels() is invoked in the next nearest non-RTT frame - where RTT texture is needed for screen rendering, in this way GPU and CPU can work simultaneously and stalls are minimized. This improves the performance significantly. It is recommended to enable 'Synchronous rendering' as well - I saw some issues with level transitions in Rez without this option enabled.

OpenGL ES 2.0 fully compatible solution was implemented, when reading pixels from a framebuffer object (FBO).
For example, when a texture (or color renderbuffer) bound to the FBO is RGBA5551 format/type, GL_IMPLEMENTATION_COLOR_READ_TYPE query is executed just before glReadPixels() to explore GPU's supported image type, when GL_UNSIGNED_SHORT_5_5_5_1 will be returned, we proceed the standard way, but if something other is returned GL_UNSIGNED_BYTE type and GL_RGBA format are used to read pixels - OpenGL ES 2.0 compatible hardware must always support this format/type. We need to properly convert it back to 16 bit color value then. This fixes an issue with ARM Mali-450 MP GPU, where (as I saw) GL_RGB/GL_UNSIGNED_SHORT_5_6_5 is the only 16 bit format/type supported to read back from a framebuffer object.

There is an issue on Adreno 506 with RGBA5551 format/type, to be specific I was able to observe a huge performance drop when using this image format/type, even though this GPU declares that it is supported (GL_IMPLEMENTATION_COLOR_READ_TYPE query). This format/type is supported, but probably there is some time expensive internal format conversion, which makes it unusable, at least here. We can see this issue very clearly in case of Rez game. Using framebuffer object with GL_UNSIGNED_BYTE texture, and then reading it using glReadPixels() with the same image format/type combination seems to be a workaround for this issue. It is strange that 32 bit color rendering works faster than RGBA5551 16 bit rendering, but it is true in this case. I have not altered TextureCacheData struct (gltex.cpp) and TexCache.h/cpp files to adapt them to support 32 bit colors, because I decided it will be a lot of changes just for this fix and because we do not need 'texconv' in this case, 32 bit image is directly written to the texture after read (bypassing TextureCacheData's 'Update()' method).

To achieve this blur effect in Rez:

screenshot_2018-12-28-14-28-11-662_com reicast emulator

blending feature is used in gldraw.cpp:
`glBlendFunc(SrcBlendGL[gp->tsp.SrcInstr],DstBlendGL[gp->tsp.DstInstr]);`
with both parameters set to GL_ONE (additive blending).
For this to work properly with different formats/types of images I needed to modify BindRTT() function code to create texture/renderbuffer/framebuffer and initialize it only once at the initialization state of a given texture address.

I have added RTT swizzle texture support too. It is used in Virtua Tennis during logo and demo video transitions to achieve 640 pixels in width of a texture.

As I saw in the source code, there is no standard code convention and a lot of the C++ code is mixed with the C code, but I have tried to keep the similar convention and limit the refactoring.

Default option for 'Render to Texture' is set to 'Zeros'. 'Full' option must be selected to enable multipass RTT.

Solution was tested with several games that use RTT (including: Rez, Virtua Tennis, THPS2 and Crazy Taxi) and is working fine.

Testing platform:
- Huawei P8 lite, Android 6 (ARM Mali-450 MP)
- Xiaomi Redmi Note 4 Snapdragon, Android 7 (Adreno 506)
- Xiaomi Pocophone F1, Android 9 (Adreno 630)

Feel free to test.

skmp:
This pull request **introduces 3 alerts** when merging 2aadb3cdd446fff8e304e995928c9cbf55391234 into 6936bc2ec503594e9fc75e687fa060426bb240a5 - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-ff11a5105bd14649f169fa5470a203cff4705286)

**new alerts:**

* 2 for Local variable hides global variable
* 1 for Multiplication result converted to larger type

---

mar753:
I have added the stencil support as well (if "OES_packed_depth_stencil" extension is available). This fixes the dimmed screen issue in Crazy Taxi.

If you would like to test this solution please fetch my branch: https://github.com/mar753/reicast-emulator/tree/multisample_rtt_implementation
because current master doesn't work on my Xiaomi devices (works only on P8 lite) - when I try to load a game I am entering the bios settings.

@flyinghead Could you take a look at this PR

baka0815:
@flyinghead were all your points worked on or is there still something left from your pov?

skmp:
This pull request **introduces 3 alerts** when merging 10d6f2526affce210baeaf822dc8e0a03b36b5c6 into 3c57177d38adecc0913aed4ac26ad726292e7e46 - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-c3f1f323f1b4eb1108adf1c1eba0402f9fe550a9)

**new alerts:**

* 2 for Local variable hides global variable
* 1 for Multiplication result converted to larger type

---

flyinghead:
This PR still reads back the generated texture from the GPU to the dreamcast VRAM, from where it is sent back to the GPU to create a new texture. In 99% of the case this is not needed: you can leave the texture on the GPU and avoid the calls to glReadPixels() and glTexImage2D() as well as the texture conversions to/from dreamcast-supported formats.

I understand that THPS2 needs this but it should be the exception rather than the rule.

There should be a an RTT option to avoid these unnecessary texture transfers and handle the general case efficiently.

mar753:
@flyinghead 
Ok, requested feature added. Render to texture for RGB565 (except for cases like THPS2) and ARGB4444 will reuse rendered texture (on GPU) without calling glReadPixels(), this should improve the performance.

mar753:
As I have checked FBOs are not supported in OpenGL 2.x without the "GL_EXT_framebuffer_object" extension, so this last commit is not needed because OpenGL 3.0 at least is required anyway.

baka0815:
@flyinghead are all your remarks addressed?

Since the default for RTT is 0 the behaviour is the same as before I have no problem merging this one.
I tested THPS and the results are as described in the PR.
I also tested Crazy Taxi and Shenmue and didn't notice any regressions (however I don't know if they use RTT).

flyinghead:
Sorry I haven't had time to dedicate to reviewing this. I assume the changes answered my concerns. 

Interestingly you mention that fb_alpha_threshold and kval_bit could change between the moment the texture is rendered and when it is used. (I must say I overlooked this case) Have you noticed this happening in some games? 
I don't have the answer but if it never happens, this would eliminate a round-trip to the GPU in all common cases (except for TPHS2 obviously).

Also I don't think Circle/All Ones and And Zeros options are needed. We need three options:
- Disabled
- Fast (cache rendered textures on the GPU)
- Full (read back pixels from GPU to VRAM then back. Needed for TPHS2 and such)
The Fast option assumes that fb_alpha_threshold and kval_bit usually don't change and that all texture types can be cached on the GPU.

mar753:
Actually I would prefer to stick with my current solution, I will explain why.
First of all option 'Disabled' is just to have a solution that is currently in the master branch (just in case of some rare situations). 'Zeros' is a must for mobile platforms to avoid artifacts like the first screen in this PR (probably not an issue in desktop version) - this option does not have an impact on performance, just like 'Ones'. 
Both 'Zeros' and 'Ones' were useful for debugging in case of games which use RTT. It is possible that in some situations 'Ones' can be useful to avoid artifacts or just to let get through some points in a game.
Dreamcast/PowerVR specification includes that fb_alpha_threshold and kval_bit must be taken into account when working with respectively ARGB1555 and KRGB1555 texture formats. I am not sure if some games actually make use of it (maybe Rez is not using it but I cannot be sure, because I am not able to test every scenario), but I am not in favor of making solution that in not entirely correct. For me the correctness is the most importang thing, only then we can work on performance improvements. Anyway, rather few games use A(K)RGB1555 format, and performance will not be spectacularly better if we keep the texture on GPU. Another counter-argument of 'Fast' option is that it will not fix the THPS2 case. For this case we have 'Shadow circle' option which was quite handy during debugging as well and most importantly it does not have (or have minimal) an impact on performance. So we can play THPS2 or similar games on older/slower mobile devices without any (or with minimal) performance drop and for sure this is a better option than no shadow at all ('Zeros').
Those options can be available, but noone enforces us to use them.

mar753:
Reicast does not handle odd resolutions, we can see that effect in case of Xiaomi POCOPHONE F1 phone where higher dimension resolution is 2159 pixels. There is a narrow line visible (Rayman 2):

line

I have fixed that in "Handle odd screen resolution (POCOPHONE fix)" commit.

I have added dynamic resolution recalculation when screen resolution changes during emulation. This fix was needed when lowering rendering resolution is active.

It's ready to be reviewed

mar753:
Moreover as I have compared on Android, native reicast (this repository) with my RTT solution from this PR (even that resolution is higher) is faster than RetroArch reicast core emulation (tested with 'Rez' - violet level)

mar753:
Added some minor fixes (e.g. to odd screen resolution handling).

Vertical scaling is now handled (SCALER_CTL), this fixes the issue in Crazy Taxi pause screen, where background is not properly aligned (see the red arrow):

mar753:
Moreover as I have compared on Android, native reicast (this repository) with my RTT solution from this PR (even that resolution is higher) is faster than RetroArch reicast core emulation (tested with 'Rez' - violet level).

This pull request is complete and ready to be merged.

mar753:
Moreover as I have compared on Android, native reicast (this repository) with my RTT solution from this PR (even that resolution is higher) is faster than RetroArch reicast core emulation (tested with 'Rez' - violet level).

@baka0815 @skmp @dmiller423 This pull request is complete and ready to be merged.

mar753:
Added some minor fixes (e.g. to odd screen resolution handling).

Vertical scaling is now handled (SCALER_CTL), this fixes the issue in Crazy Taxi pause screen, where background is not properly aligned (see the red arrow):

mar753:
Moreover as I have compared on Android, native reicast (this repository) with my RTT solution from this PR (even that resolution is higher) is faster than RetroArch reicast core emulation (tested with 'Rez' - violet level).

@baka0815 @skmp @dmiller423 This pull request is complete and ready to be merged.

core/rend/gles/gles.cpp
```diff
@@ -1964,61 +1979,81 @@ bool RenderFrame()
...
-		GLuint channels,format;
+		const char * renderer = (char *)glGetString(GL_RENDERER);
+		const char * adreno506Renderer = "Adreno (TM) 506";
+		GLuint channels, format;
 		switch(FB_W_CTRL.fb_packmode)
 		{
 		case 0: //0x0   0555 KRGB 16 bit  (default)	Bit 15 is the value of fb_kval[7].
 			channels=GL_RGBA;
-			format=GL_UNSIGNED_SHORT_5_5_5_1;
+			// When using RGBA5551 format on Adreno 506, rendering is extremely slow
```
skmp: Keep in mind, powervr will always do 32bpp rendering. The frame-buffer format only when copying to memory from the internal tile storage.

core/rend/gles/gles.cpp
```diff
@@ -1964,61 +1979,81 @@ bool RenderFrame()
 
 		ShaderUniforms.Set(s);
 	}
+
 	//setup render target first
 	if (is_rtt)
 	{
-		GLuint channels,format;
+		const char * renderer = (char *)glGetString(GL_RENDERER);
```
skmp: Wouldn't this be better in gl_init? Then we can do like `gl.workarounds.<whatever>`

core/rend/gles/gltex.cpp
```diff
@@ -364,119 +394,496 @@ struct TextureCacheData
...
 typedef map<u64,TextureCacheData>::iterator TexCacheIter;
 
 //TexCacheList<TextureCacheData> TexCache;
 
 struct FBT
 {
-	u32 TexAddr;
-	GLuint depthb,stencilb;
+	u32 texAddress;
+	u16 texData[1024*1024];
```
skmp: Can we avoid this by writing directly to vram on readouts?

core/rend/gles/gltex.cpp
```diff
@@ -20,11 +27,17 @@ Compression
...
+const u32 shadowCircleH = 128;
+u16 shadowCircleTexture[shadowCircleW][shadowCircleH] = {0};
+set<u32> delayedUpdateQueue;
+
 #if FEAT_HAS_SOFTREND
 	#include <xmmintrin.h>
 #endif
 
 u16 temp_tex_buffer[1024*1024];
+u32 temp_tex_buffer_32[1024*1024];
```
skmp: Can `temp_tex_buffer` be resized and used here to avoid allocating 4 extra megs of memory ?

core/rend/gles/gltex.cpp
```diff
@@ -260,8 +273,12 @@ struct TextureCacheData
 		}
 	}
 
-	void Update()
+	void Update(bool isSourceVram = true, u16 *sourceData = NULL)
```
skmp: Writing directly to vram and avoiding `FBT::texData` would make  `isSourceVram` and `sourceData` wouldn't it?

core/rend/gles/gles.cpp
mar753: Ok, but at the end it will be 16bpp anyway for RTT? This solution uses 16bpp rendering from the beginning because it should be faster in most cases (for slower mobile GPUs) than 32bpp rendering, should something be changed?

core/rend/gles/gles.cpp
mar753: Yes, sure I will move it to gl_init

core/rend/gles/gltex.cpp
mar753: Actually, to improve performance RTT is implemented in a way that vram is not used (because vram is not valid in the next frame), so I need to store a texture for future use. I wrote about it in this comment (third paragraph): https://github.com/reicast/reicast-emulator/pull/1460#issuecomment-439205071

core/rend/gles/gltex.cpp
mar753: It is easier to read and less error prone when we separate those buffers (one is for 16bpp color formats and second is for 32bpp color format). 4MB is rather small amount of memory, considering that smartphones have even 8GB of RAM nowdays

core/rend/gles/gltex.cpp
mar753: It is added for the same reason as above -> to improve performance RTT is implemented in a way that vram is omitted (because vram is not valid in the next frame), why is that I wrote in this comment (third paragraph): https://github.com/reicast/reicast-emulator/pull/1460#issuecomment-439205071

mar753:
@skmp  Find my answers below

> @mar753 did a really quick scroll though the diff and added some comments.
> 
> In future PRs, would be great to
> 
> * avoid nonfunctional whitespace changes to make reviewing easier. Trying to understand what was actually changed in `gl_tex.cpp` was a bit of a challenge.

You are right, I prefer spaces instead of tabs (which looks strange on GitHub page) in source files, but just to be consistent with the rest of files I have used tabs to indent source files.

> * Split changes that are not directly related to the main PR purpose into separate PRs (eg odd-sized-rendering workaround)
> 

I agree to this in future PRs, just though that changes like odd screen resolution are so small that I do not need to create separate PRs.

> Here's a couple more questions
> 
> * Are textures  always read back to the host ram and then re-uploaded?
> * How does this compare with @flyinghead RTT changes / `fh/mymaster`? There's quite some divergence from `master`.
> 

I will briefly write how it works. 
So in 'fh/mymaster' solution when RTT frame is processed, rendering and glReadPixels() are invoked within the same frame processing cycle (one by one). As we know glReadPixels() is a blocking operation, so we will freeze CPU on this call until rendering will be completed and GPU will return pixel data. I have tested it and this solution can cause framerate drops on slower/mid range portable devices.
In my implementation, when RTT frame is processed I start rendering to a texture, but I do not invoke glReadPixels() right after that, it will be invoked in the the next frame processing cycle. This way, when rendering is executed on GPU we can execute code on CPU not wasting precious clock cycles. That's why there is some additional code and variables needed. What is more those stored variables can be reused when gl_GetTexture() is called, until they will be invalidated by the next RTT rendering cycle (with the same address). Moreover, in the recent version of my RTT implementaion, ARGB4444 and RGB565 are stored in the OpenGL's video memory, and when they are needed we will just pass a texture ID and glReadPixels() is not needed at all (other formats like KRGB1555 need it because OpenGL hardware cannot substitute 'K' with a given 'fb_kval[7]' value, this must be done on CPU and for that we need glReadPixels() call).

> I think it's up to @flyinghead to have final say on this. If it doesn't upset the code too much otherwise, I'd merge it.

mar753:
@skmp
Requested changes added, next steps are yours.

mar753:
@flyinghead Can you approve this PR?

mar753:
@baka0815 @dmiller423 @skmp 
PR is accepted, thus we have a green light to merge this solution.

baka0815:
Since @skmp is ok with this and @flyinghead approved it, I'm merging this.

Thanks @mar753 for your work!

PR #1479, opened 2019-01-07
Lowering rendering resolution option

mar753:
When I was testing my additional pull request implementing RTT: https://github.com/reicast/reicast-emulator/pull/1460 I saw that Adreno 506 has performance issues when rendering using native resolution (1920x1080). I have observed framerate drops in THPS2, so I came with an idea that we should be able to lower the rendering resolution, thus improve the performance.

I have implemented a solution where we can decrease vertical and horizontal resolution (%) in emulator experimental options section.
How it works?
Instead of rendering to the default framebuffer, we render the whole screen to a texture (with native resolution size) but we set a lower viewport value (depending on vertical/horizontal resolution options). Then we use this texture and render it to the default framebuffer, extending it to cover the full screen. Then we apply scissor settings and draw OSD (it is drawn at the full resolution to keep the highest quality). When vertical/horizontal resolutions are set to 100% (default value), the current solution (what is in master branch) is used - so there is no impact on the performance.

Here is an example when full native resolution is used (currently in master branch), vertical/horizontal resolution set to 100% (1920x1080):
screenshot_2019-01-07-22-12-01-060_com reicast emulator

And here is the example where vertical and horizontal resolutions were set to 25% of the native screen resolution (about: 480x270):
screenshot_2019-01-07-22-10-37-140_com reicast emulator

If you would like to test how this solution works, please use my branch: 
https://github.com/mar753/reicast-emulator/tree/lowering_rendering_resolution_option
because current master doesn't work on my Xiaomi Pocophone/Redmi 4 Note devices (works only on Huawei P8 lite) - when I try to load a game I am entering the bios settings.

Please review my second PR as well, where multipass render to texture was implemented: https://github.com/reicast/reicast-emulator/pull/1460
Could you create a test branch for me in this repository as well and connect it to CI?

### DIFF
--- a/core/rend/gles/gldraw.cpp
+++ b/core/rend/gles/gldraw.cpp
@@ -1081,7 +1081,7 @@ void DrawModVols()
 		glStencilFunc(GL_EQUAL,0x81,0x81); //only pixels that are Modvol enabled, and in area 1
 		
 		//clear the stencil result bit
-		glStencilMask(0x3);    //write to lsb 
+		glStencilMask(0x3);    //write to lsb
 		glStencilOp(GL_ZERO,GL_ZERO,GL_ZERO);
 #ifndef NO_STENCIL_WORKAROUND
 		//looks like a driver bug ?
--- a/core/rend/gles/gldraw.cpp
+++ b/core/rend/gles/gldraw.cpp
@@ -1155,3 +1155,89 @@ void DrawStrips()
 #endif
 	}
 }
+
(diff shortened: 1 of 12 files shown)


skmp:
@mar753 I'm a bit lost with all the recent merging. Is this now available on master via some other MR?

mar753:
@skmp Yes, this one: https://github.com/reicast/reicast-emulator/pull/1480

PR #1485, opened 2019-01-16
ALSA: configurable output device

### LINKED ISSUE #1483: ALSA: configurable output device
In ``/core/oslib/audiobackend_alsa.cpp`` we're using the ``default`` device and if that's not available we're trying ``plughw:0,0,0`` and ``plughw:0,0``.
https://github.com/reicast/reicast-emulator/blob/3c57177d38adecc0913aed4ac26ad726292e7e46/core/oslib/audiobackend_alsa.cpp#L20

In my case the device I want to use is ``plughw:1,0``(because 0,0 is the HDMI audio out from the GPU) and so I'm getting a pcm error and reicast ist dying.

We should create an ``[alsa]`` section in the ``emu.cfg`` with a ``device=`` entryto specify the device.

baka0815:
Trying to load the device from the ``~/.config/reicast/emu.cfg`` section ``alsa`` and if not set (should be the default for everyone currently) use the existing procedure to try to determine the device (``default`` > ``plughw:0,0,0`` > ``plughw:0,0``).
If the init was successfull write the result back to the configuration file to make the next start a bit faster.
If the device is set then just try to use this as the ALSA device.

Also removes two unused variables while we're at it.

Closes #1483

### DIFF
--- a/core/oslib/audiobackend_alsa.cpp
+++ b/core/oslib/audiobackend_alsa.cpp
@@ -8,10 +8,6 @@ snd_pcm_t *handle;
 // We're making these functions static - there's no need to pollute the global namespace
 static void alsa_init()
 {
-
-	long loops;
-	int size;
-
 	snd_pcm_hw_params_t *params;
 	unsigned int val;
 	int dir=-1;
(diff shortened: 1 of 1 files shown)


baka0815:
I no one objects, I'm going to merge this today.

Issue #1371, opened 2018-09-03
Make SDL input usable


baka0815:
The SDL input backend currently is hardcoded (like the X11 keyboard backend), caputures the mouse and only works with the first joystick.
The following tasks should be done to make the experience much nicer:

- [x] Make keyboard work (#1370)
- [ ] Configurable key/button layout
- [ ] Allow more than one joystick
- [ ] Configurable maple devices (VMU, PurupuruPack, Mouse)
- [ ] Use [GameController](https://wiki.libsdl.org/CategoryGameController) API
- [x] Rumble support (#1486)
- [ ] Split SDL GUI and SDL INPUT support
- [ ] Support hot-plugging work (support SDL_JOYDEVICEADDED and SDL_JOYDEVICEREMOVED)

baka0815:
@flyinghead could you take a look what of the above list is done already since we merged your branch?

flyinghead:
Everything is done except splitting GUI and input code.

baka0815:
Ok, since @dmiller423 or @MrPsyMan said that it's unwanted to split GUI and input code, we can close this.

Issue #340, opened 2014-02-25
Maple: Support jump pack


hooby3dfx:
http://en.wikipedia.org/wiki/Dreamcast_accessories#Jump_Pack

I GOT THIS

Holzhaus:
We should probably split this task up since it there are basically two things to do:
- Get rumble/jumppack support at the emulator side and write it into a variable (like these `kcode`, `joyx`, `joyy`, etc. variables) or using a callback function. This is something I can't do, since I don't have in-depth knowledge about dreamcast internals.
- Handle the rumble events on the OS side. I'd volunteer and implement this on Linux (evdev).

skmp:
Yeh, it needs a new callback. Ill look on the emulation side when i have
some more time, it is trivial to add
Am 24.08.2015 13:50 schrieb "Jan Holthuis" <notifications@github.com>:

Holzhaus:
Just stumbled upon this, could also be helpful:
http://sourceforge.net/p/cadcdev/kallistios/ci/d9c299e747991c7ba85a6c5811ac4b226b481eb0/tree/kernel/arch/dreamcast/include/dc/maple/purupuru.h

baka0815:
This should be fixed by now and can be closed?

AbandonedCart:
Completely missing on Android

skmp:
@hooby3dfx I assume you no longer plan to work on this?

baka0815:
It's also not implemented for SDL afaik (#1371).

baka0815:
It **is** implemented in SDL code. So only backend left is Android?

skmp:
Should be implemented after #1523

PR #1486, opened 2019-01-16
SDL: rumble support

baka0815:
This adds support for rumble to the SDL backend.
It does not however attach a purupuru/rumble pack.

This is one part to make the SDL input backend on par with the evdev backend (#1371, #340).
After that work is done, I think we can work on a more unified input backend implementation like the audiobackend (see #843, #823).


baka0815:
I no one objects, I'm going to merge this today.

PR #1533, opened 2019-03-30
win32: resize/maximize window

flyinghead:
save size and state on exit
add app icon and version resource
hide console by default. -console option to enable it. -log <file> to log to a file
visual studio partial build compat


core/core.mk
```diff
@@ -5,7 +5,7 @@
 #LDFLAGS	:= -Wl,-Map,$(notdir $@).map,--gc-sections -Wl,-O3 -Wl,--sort-common
 
 RZDCY_SRC_DIR ?= $(call my-dir)
-VERSION_SRC := $(RZDCY_SRC_DIR)/version/version.cpp
```
skmp: Doesn't this make rebuilds a HUGE pain? Every checkout forces one

core/archive/archive.cpp
```diff
@@ -21,12 +21,15 @@
 
 #include "archive.h"
 #include "7zArchive.h"
+#ifndef _MSC_VER
```
skmp: Why do we need the check here?

skmp: Oh, existing VS build is broken.

PR #1540, opened 2019-04-04
GL renderer optimizations for low-end devices

flyinghead:
New option to disable fog effects
Fix GL state cache that was always disabled
put all shaders in a map and set uniforms once each frame
Use the GL cache in imgui, avoid saving/restoring state

### DIFF
--- a/core/nullDC.cpp
+++ b/core/nullDC.cpp
@@ -513,6 +513,7 @@ void InitSettings()
 	settings.rend.CustomTextures    = false;
 	settings.rend.DumpTextures      = false;
 	settings.rend.ScreenScaling     = 100;
+	settings.rend.Fog				= true;
 
 	settings.pvr.ta_skip			= 0;
 	settings.pvr.rend				= 0;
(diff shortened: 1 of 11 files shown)


flyinghead:
- The problem is the number of entries in the table, due to new options that weren't previously implemented (trilinear filtering, bump maps, gouraud shading and color clamping). That's around 25,000 combinations so big table and slow iterations. Using a hashmap doesn't seem to hurt on slow devices.

- Completely disabling the fog is a requested feature (when using custom HD textures). It also helps slow devices. The previous exponential matching was never active iirc. Not sure it is worth a third option.

skmp:
True, but we don't need all 25K shaders alive at once, we can just flip the table on options change. Or clear the cache.

Not all platforms support runtime shader compilation, so it's nice to have a limited set of active shaders for porting. This has crept up a few times before (xbox, xbox360, ps3, xbone, winrt, winphone, and others)

There are 2-3 standard fog features used by games, either EXP fog or LIN fog. The maths-based implementation can get it 98% correct at almost no speed hit.

Historically, clx2 uses a lookup table to avoid doing exp() in hardware, as it would be too slow/complex.

Emulatorer:
Didn't feel any improvement using latest beta build of reicast game(s) are still laggy compare to the previous build without fh/master build :(

skmp:
@Emulatorer are you on http://chat.reicast.com? We need to bisect this a bit I think.

PR #1541, opened 2019-04-05
Avoid opposite dpad buttons being both down. Ignore android key repeats

flyinghead:


core/input/gamepad_device.cpp
```diff
@@ -51,7 +51,39 @@ bool GamepadDevice::gamepad_btn_input(u32 code, bool pressed)
 	if (key < 0x10000)
 	{
 		if (pressed)
+		{
 			kcode[_maple_port] &= ~(u16)key;
```
skmp: This is probably best implemented at the maple side, as it looks like a feature of the dreamcast gamepad that all backends need to follow

flyinghead: Right, not to mention it uses the Dreamcast key codes, which are different for Naomi or Atomiswave. However doing it here allows one to process the button down event (and cancelling the opposite down state). Doing it on the maple side would imply ignoring the button down state for both buttons if both are down.
I'll add a #ifdef around this code to restrict it to dreamcast mode. I still have to wire the Naomi and Atomiswave inputs so this code will change.

PR #1549, opened 2019-04-10
Remove trailing whites-spaces ins sourcecode files

baka0815:
This only removes the white-spaces at the end of lines in our source files (*.c, *.cpp, *.h). I omitted the whole ``deps`` folder.


flyinghead:
Whaaaat? Please, don't

baka0815:
@flyinghead why not?

flyinghead:
Because it makes reading git history much harder, breaks git blame, and all the reasons why we don't want whitespace commits

skmp:
I think making a single merge that fixes the issue once and for all is OK, it is easy to skip it on blame and such. Having it creep in in unrelated commits i find very disruptive though

skmp:
Since this touches pretty much all files, i think it makes sense to also address the formatting issues in the same PR. @baka0815 you on discord so we can discuss a bit?

skmp:
(and make CI enforce some rules, and provide rule files for popular IDEs)

baka0815:
@skmp I had a clang format file in #1311 but there were still some discussion points open.
I reopened the PR for the formatting configuration file to discuss that there.

skmp:
Will complete the review after the Orthodox Easter

skmp:
@baka0815 let's close this and address it after the refactor PR is done

PR #1568, opened 2019-04-24
AUDIO: Allow backend specific settings

baka0815:
The settings can be defined by each audio backend (currently only alsa is implemented) and of different types (integer with min and max value, checkbox and a list with a callback).

I think this is now ready for the moment as any further work can be done by subsequent prs.

alsa device list

### DIFF
--- a/core/oslib/audiobackend_alsa.cpp
+++ b/core/oslib/audiobackend_alsa.cpp
@@ -175,12 +175,81 @@ static void alsa_term()
 	snd_pcm_close(handle);
 }
 
+std::vector<std::string> alsa_get_devicelist()
+{
+	std::vector<std::string> result;
+
+	char **hints;
+	int err = snd_device_name_hint(-1, "pcm", (void***)&hints);
+
+	// Error initializing ALSA
+	if (err != 0)
+		return result;
+
+
+	char** n = hints;
+	while (*n != NULL)
+	{
+		// Get the type (NULL/Input/Output)
+		char *type = snd_device_name_get_hint(*n, "IOID");
+		char *name = snd_device_name_get_hint(*n, "NAME");
+
+		if (name != NULL)
+		{
+			// We only want output or special devices (like "default" or "pulse")
+			// TODO Only those with type == NULL?
+			if (type == NULL || strcmp(type, "Output") == 0)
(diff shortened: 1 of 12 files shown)


skmp:
Will complete the review after the Orthodox easter

skmp:
I’d re-order the FPS, DSP and maybe add CDDA on top of the backend/device selection, as they are not backend specific options

core/oslib/audiobackend_alsa.cpp
```diff
@@ -175,12 +175,81 @@ static void alsa_term()
 	snd_pcm_close(handle);
 }
 
+std::vector<std::string> alsa_get_devicelist()
```
skmp: I think you can just use `vector<string>` here to avoid cramping your hands. Afair reicast has `<vector>`, `<string>` and `using std;` on the main include file (`types.h`)

skmp:
Overall, LGTM. See comments

core/oslib/audiobackend_alsa.cpp
baka0815: I think I will leave this as is, as I already typed this. :-)
I will remember this for the future and omit the ``std::``.

baka0815:
> I’d re-order the FPS, DSP and maybe add CDDA on top of the backend/device selection, as they are not backend specific options

Ok, I reordered the settings, but I don't know what CDDA option I should put there?

core/oslib/audiobackend_alsa.cpp
```diff
@@ -97,7 +110,7 @@ static void alsa_init()
 	rc=snd_pcm_hw_params_set_rate_near(handle, params, &val, &dir);
```
baka0815: Should the sample rate be configurable or is 44.100Hz needed for something?

skmp: 44.1 is needed

core/oslib/audiobackend_alsa.cpp
skmp: well, to keep in mind for the future

baka0815: Yep, I will.

PR #1569, opened 2019-04-25
PS4 on master,  many build fixes for subtargets, ability to remove save states and gles linkage that can fail silently - unfinished

dmiller423:
Hopefully this does a proper diff to see changes


dmiller423:
I'll just leave this as a branch,   one can pull the guards used out or view them for reference.
It contains some further debug printfs that should not be merged.

PR #1570, opened 2019-04-26
Fix cmake for Linux builds & Disable Modem option

davidgfnet:


core/hw/holly/sb_mem.cpp
```diff
@@ -217,6 +217,8 @@ T DYNACALL ReadMem_area0(u32 addr)
 	{
 #if DC_PLATFORM == DC_PLATFORM_NAOMI || DC_PLATFORM == DC_PLATFORM_ATOMISWAVE
 		return (T)libExtDevice_ReadMem_A0_006(addr, sz);
+#elif defined(DISABLE_MODEM)
```
baka0815: Why don't we use a 'ENABLE_MODEM' switch for a possitive phrasing?

ZirconiumX: Because the modem is enabled by default. I don't think positivity of phrasing matters any.

davidgfnet:
Lemme close this since cmake changes are irrelevant it seems :dancer:

PR #1571, opened 2019-04-27
Add ENABLE_MODEM and rework makefiles to support it.

davidgfnet:


core/hw/holly/sb.cpp
```diff
@@ -783,7 +783,7 @@ void sb_Init()
 	maple_Init();
 	aica_sb_Init();
 
-#if DC_PLATFORM == DC_PLATFORM_DREAMCAST
+#if DC_PLATFORM == DC_PLATFORM_DREAMCAST && defined(ENABLE_MODEM)
 	ModemInit();
```
skmp: I’d rather us `#define ENABLE_MODEM 0/1`, using `defined(...)` often leads to typos

davidgfnet:
Although there was some pushback on discord regarding this and how it works with cmake support and all, I'm merging this since there's enough consensus. Will be happy to follow up with another PR to fix things that might be missing here.

PR #1572, opened 2019-04-27
Improve stdclass.h/cpp support for non-Linux non-Windows platforms

davidgfnet:
Simplifies having implementation on platform separated files, which is a
pain for platforms which are not Windows but not Linux either (and yet
support pthreads).
Some minor cleanup here and there while we are at it.


dmiller423:
It really just moves some things around,  and thread/mutex/* were meant to be moved to their own platform respective files on my cmake branch as well with a posix-common base.  The entire platform system really needs an overhaul.   Stuffing it into stdclass with a bunch of ifdef's ...  does it help anything?   ;shrug;

I'll  OK it but it doesn't solve any problems,  you have checked what removing the default constructor arguments does on all platforms?

skmp:
Looks good to me overall

skmp:
Plus, we can most likely provide a default, stdlib based, implementation there

davidgfnet:
@dmiller423 I'm 100% in agreement with what you say. In the meantime though, this enables platforms that are not very Linux-like (I guess PS4 might be also benefiting from this?) like the Switch.
I also find it a bit odd to define the headers in a thing like stdclass and the implementation on winmain. Perhaps we should have a proper lib to hide all these details. I'm 100% for it, but in the meantime let's at least keep the ugliness and creep in some logical place and related file :)
Thanks for the feedback!

davidgfnet:
Also note this solves a potential issue: if sizeof(pthread_t) > sizeof(void*) the code does not work. This fixes it. Luckily we never hit that cause in most platforms pthread_t is an integer :D (Or at least it is made opaque to the user by exposing an int as storage type)

PR #1577, opened 2019-05-01
audiostream only wait if at speed

dmiller423:


core/oslib/audiostream.cpp
```diff
@@ -117,10 +117,14 @@ audiobackend_t* GetAudioBackend(std::string slug)
 	return NULL;
 }
 
+extern double full_rps;
+
 u32 PushAudio(void* frame, u32 amt, bool wait)
 {
+    bool do_wait = (full_rps<50.f)?false:wait;
```
baka0815: Your indentation is broken here (spaces instead of tab(s)).
```suggestion
	bool do_wait = (full_rps < 50.f) ? false : wait;
```

PR #1578, opened 2019-05-01
Win32 fullscreen via alt-enter

dmiller423:


core/windows/winmain.cpp
```diff
@@ -331,6 +335,14 @@ LRESULT CALLBACK WndProc2(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
 			keyboard.keyboard_input(keycode, message == WM_KEYDOWN);
 		}
 		break;
+	
+	case WM_SYSKEYDOWN:
+		if (wParam == VK_RETURN)
+			if ((HIWORD(lParam) & KF_ALTDOWN))
```
gigaherz: Nitpick: Couldn't those two IFs be merged into one?

dmiller423: yes i'm sure they could

dmiller423: we're just going to chop this one up to readability is better w. two , and let travis get some idle cycles in...

PR #1585, opened 2019-05-08
Add SDL2 audio backend.

davidgfnet:
The backend supports 44.1KHz and 48KHz (with resamping). The resampler
is not great, has some noise but no idea where it comes from.
This enables the switch port, since using SDL2 is the quickest way
to get audio working.
TODO: Add support in the cmake, once cmake is fixed at master/HEAD.


davidgfnet:
Perhaps you can play with it a bit later (just enable it in the makefile and then force it to assume 44.1 is not available so it falls back to 48khz) and see whether you can improve the audio quality :)

PR #1594, opened 2019-05-10
Fix TARGET_NO_NVMEM and deprecate TARGET_NO_EXCEPTIONS

davidgfnet:
Linked them both toghether since you can't really define one and
not the other (plus Linux honors one windows the other in some
cases).
More refactoring on this area to follow.


davidgfnet:
Dropping this in favour of a new PR

PR #1595, opened 2019-05-10
Fix Modem & serialization regression

davidgfnet:
This should fix serialization across platform (plus with previous savestates).
It will also re-enable modem support in OSX and Android.


davidgfnet:
Works according to my testing, will merge.

Issue #1592, opened 2019-05-10
Reicast beta  regression on arm32 android since flyinghead branch merge to master


Emulatorer:
**Please Note: This form is the minimum required information for submitting bugs.**  
**Removing this form may lead to your issue being closed until it is completed.**

| Platform | Branch | Hash | CIDL |
| Android | Playstore | ---- | ---- |
| Installed OS and version (ie. Android 8.0) | Branch (or Google Play) | Hashtag (skip if unknown) | Downloaded from build site? |

**_Description of the Issue_**
All my reicast game are running slow since https://github.com/reicast/reicast-emulator/pull/1523 but running smooth using r8.1-playstore

**_Log Gathered_**
Maybe this log could help
[reicast_beta_log_.txt](https://github.com/reicast/reicast-emulator/files/3165211/reicast_beta_log_.txt)
[reicast_r8.1_log.txt](https://github.com/reicast/reicast-emulator/files/3165216/reicast_r8.1_log.txt)

  * Fill in any steps already tried here
  * Begin each new line with an asterisk
  * If no steps are required, please skip
```

Please paste the contents of the log / logs here
You may leave this field blank if you have none.

```

**_Screenshots_**

(Replace this line with any issue screenshots)

- [X ] Place an X in the box to confirm you have filled in this form

Emulatorer:
@flyinghead you have a phone with arm32 like Mali-400 GPU right? It is running slow in your phone using latest beta build?

flyinghead:
I do. I have a device with an arm32 with Mali-400 GPU running Android 4.4.4.
I also test with a Raspberry Pi.

I don't see a significant slowdown between now and before the merge.

Emulatorer:
@flyinghead Well it's running slow on my device with Mali-450 MP GPU arm32 :(
Here's the video sample
https://drive.google.com/file/d/1A-p0o-YVhsgin9xm-dbA6YXf1xChtQkh/view?usp=drivesdk
The first video Im using is beta from playstore and the second video is the r8.1 from playstore.

You can see the difference of the speed and fps of the game I tested it in MVC2 but other my games in reicast are performing poor using latest beta started from your branch the previous beta are fine :/

kage2051:
You could try using a lower value for scaling, a value of 100 for it means native resolution of your device. For example, on my SD625 (ARM64) tablet with 1080p screen, a scale of 70% means the games will run at ~800p on that device. This also improves speed in most devices with older or low-end processors.

[...]

PR #1603, opened 2019-05-15
Fix build errors on GL ES 2.0 platforms (RPi...)

flyinghead:


Emulatorer:
@flyinghead possible improvement for #1592?

PR #1606, opened 2019-05-16
Fixes for cmake + msvc

davediff:
Also now builds .asm file for dynarec jit with msvc


skmp:
Just as an overall architecture point, I’d rather us move away from .asm files as they are not portable. In the bast, the best solution has been to reuse the runtime emitter to generate the helpers on startup- though that /can/ add a small overhead

davediff:
@skmp Agreed! @davidgfnet is looking at doing that exact thing and removing the .asm file.

skmp:
👌

PR #1613, opened 2019-05-20
 audio: new option for automatic limitFPS 

flyinghead:
- LimitFPS now can be enabled, disabled or automatic. The latter option doesn't wait for audio if the emulator is running slow (over the last 2 secs)

- Renamed `Audio Backend` to `Audio Driver` in the UI.

- alsa: don't spam console if underrun


baka0815:
Looks good to me! 👍 
However I'm unsure if we want to completely remove the prints in the ALSA backend or if we want to use one of the debug print directives.
I'm also not sure if driver will be more understandable than backend.

skmp:
+1 for debug prints

PR #1617, opened 2019-05-24
Cleanup and unify x64 asm implementation

davidgfnet:


core/hw/sh4/dyna/driver.cpp
```diff
@@ -38,6 +38,7 @@ uintptr_t cc_rx_offset;
 u32 LastAddr;
 u32 LastAddr_min;
 u32* emit_ptr=0;
+t_mainloop_fn mainloop_function;
```
skmp: I'd name this `MainLoopFN_t` or some such to fit with the original nullDC naming

skmp: `MainloopFp_t* whatever`

core/hw/sh4/dyna/driver.cpp
```diff
@@ -88,7 +89,7 @@ void recSh4_Run()
 		printf("Warning: SMC check mode is %d\n", settings.dynarec.SmcCheckLevel);
 	
 	verify(rcb_noffs(&next_pc)==-184);
-	ngen_mainloop(sh4_dyna_rcb);
+	(*mainloop_function)(sh4_dyna_rcb);
```
skmp: Do you need to `*` here? I think just `mainloop_function()` works

core/hw/sh4/dyna/ngen.h
```diff
@@ -95,7 +95,8 @@ u32 DYNACALL rdv_DoInterrupts_pc(u32 pc);
 
 //Stuff to be implemented per dynarec core
 
-void ngen_init();
+typedef void (*t_mainloop_fn) (void*);
+t_mainloop_fn ngen_init();
```
skmp: Ideally we would return an `ngen_interface_t` here. Or passing it as an `ngen_interface_t* if` out parameter

PR #1619, opened 2019-05-24
Add more targets to Wercker, including Windows

davidgfnet:
Upgrade mutexes to cMutex to avoid issues on Windows
Seems that we cannot include stdclass.h from a header without
things going bad, due to some cyclic dependency in preprocessed
stuff?

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -4,10 +4,16 @@ build:
     steps:
         - script:
             name: install-dependencies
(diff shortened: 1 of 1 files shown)


davediff:
Should this commit only change the wrecker.yml? Looks like there are some changes to other files from your other pull requests.

davidgfnet:
Yes you are right, the other changes are just temporary, until PR #1623 is merged. Then I will drop all of them but the yaml file. Thanks!

PR #1623, opened 2019-05-25
Stop using std::mutex on Windows builds.

davidgfnet:
Reworked the threading code to an oslib and fixed some use of it.
Had to add windows.h includes and some other minor refactor since
some file were relaying on that header being already included.
While we are at it replace min/max functions with std:: ones.
Seems that we were using the windef.h defined macros (and no idea
where they come from in Linux, which sounds scary to me).


core/oslib/posix/threading.cpp
```diff
@@ -0,0 +1,105 @@
...
+	pthread_mutex_init(&mutx, NULL);
+	pthread_cond_init(&cond, NULL);
+}
+cResetEvent::~cResetEvent() {
+}
+void cResetEvent::Set()//Signal
+{
+	pthread_mutex_lock( &mutx );
+	state=true;
+    pthread_cond_signal( &cond);
```
davediff: nit: align

core/hw/maple/maple_devs.cpp
```diff
@@ -1125,18 +1125,18 @@ struct maple_sega_purupuru : maple_base
 					INC = 0;
 				bool CNT = VIBSET & 1;
 
-				float power = min((POW_POS + POW_NEG) / 7.0, 1.0);
```
skmp: In general, we have the habit of importing `std` for common functions like `min`. Too lazy to type`std::bs` all the time

davidgfnet:
Just wanna avoid using some random min/max macro that depends on some windows header, which is fragile if a header changes due to some refactor. We can later decide to use "using std::min/max" and be done with it for easy reading. Thanks!

shell/linux/Makefile
```diff
@@ -251,12 +251,10 @@ else ifneq (,$(findstring vero4k,$(platform)))
 # Windows
 else ifneq (,$(findstring win32,$(platform)))
     NOT_ARM := 1
-    CFLAGS += -DTARGET_NO_WEBUI -fno-builtin-sqrtf -funroll-loops -DHAVE_FSEEKO -D TARGET_NO_AREC
+    CFLAGS += -DNOMINMAX -DTARGET_NO_WEBUI -fno-builtin-sqrtf -funroll-loops -DHAVE_FSEEKO -D TARGET_NO_AREC
```
davediff: Also please add to cmake build.

PR #1634, opened 2019-06-11
Add social tab, buttons, cta

skmp:
Quick and dirty social tab, discord CTA in games list

Windows and android haven't been tested, hopefully they build.

@davidgfnet can we handle this in switch too?

reicast-cta-draft
reicast-social-draft

### DIFF
--- a/core/windows/winmain.cpp
+++ b/core/windows/winmain.cpp
@@ -15,6 +15,8 @@
 #include "hw/maple/maple_cfg.h"
 #pragma comment(lib, "XInput9_1_0.lib")
 
+#include <process.h>
+
 PCHAR*
 	CommandLineToArgvA(
 	PCHAR CmdLine,
@@ -810,3 +812,9 @@ void os_DoEvents()
 
 int get_mic_data(u8* buffer) { return 0; }
 int push_vmu_screen(u8* buffer) { return 0; }
+
+void os_LaunchFromURL(const string& url)
+{
+	auto cmd = "start " + url; 
+	system(cmd.c_str());
+}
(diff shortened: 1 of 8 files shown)


gigaherz:
Hmmm, I feel like it would look better if that tab was called "Help" or such? I tend to actively ignore things called "social".

core/windows/winmain.cpp
```diff
@@ -810,3 +812,9 @@ void os_DoEvents()
 
 int get_mic_data(u8* buffer) { return 0; }
 int push_vmu_screen(u8* buffer) { return 0; }
+
+void os_LaunchFromURL(const string& url)
+{
+	auto cmd = "start " + url; 
+	system(cmd.c_str());
```
gigaherz: Ew. There's windows APIs for that, which don't involve the "start" program :P
e.g. 
```c
ShellExecuteA(0, 0, url.c_str(), 0, 0 , SW_SHOW );
```

skmp: Mnn, can you implement it this way and append on the PR? I don't have a windows env setup, so this was 100% guesswork

gigaherz: Done.

skmp:
It is FB, discord, etc - that's why I named it social (for now). It's fine if people ignore it, for now, I guess :p

core/windows/winmain.cpp
skmp: Thx. Does it work as expected on windows?

gigaherz: Yup! Executing the "open" verb of an URL will open it in the default browser, same as if you were opening a document.

PR #1597, opened 2019-05-11
Added support for pull-style audiostream, including resampling pulls.

gigaherz:
Only dsound and sdl2 implemented as pull style, and the dsound implementation is rather messy.

PR #1639, opened 2019-06-15
fix make problem with pi

mpcore-hub:
fix make problems with orange pi/banana pi/tritium h3

look@
https://github.com/reicast/reicast-emulator/pull/1597


skmp:
Thanks for the PRs :+1:

PR #1708, opened 2019-07-29
Reshuffle lua loading into a modpack loading system, separate from th…

gigaherz:
…e custom texture system.


skmp:
Overall looks really good. 

Minor comments
- I'm trying to standardize to `Init`/`Reset`/`Term` and `modname_ThisKindOfCase()`
- Most likely best to move lua/customtexture to their own folders so they can be modularized
- `SCRIPTING` is probably best implemented as `MODS_NO_LUA` -- and enabled by default

skmp:
@gigaherz do you have a moment to rebase and oopify this? I have in my queue atm but i'm quite overloaded with everything else.

PR #1710, opened 2019-07-30
Fix naked functions & stack overflow

skmp:
12 hours of debugging to find this one. I'm rusty.

### DIFF
--- a/libswirl/jit/backend/arm32/rec_arm.cpp
+++ b/libswirl/jit/backend/arm32/rec_arm.cpp
@@ -317,14 +317,6 @@ struct arm_reg_alloc: RegAlloc<eReg,eFSReg,false>
 arm_reg_alloc reg;
 
 
-#ifdef naked
-#undef naked
-#define naked __attribute__((naked))
-#endif
-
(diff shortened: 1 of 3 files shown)


skmp:
@flyinghead you might want to save yourself some headaches and backport this to flycast

flyinghead:
@skmp thanks for the tip

However I don't see any prologue generated in `do_sqw_nommu_area_3` even without the naked specifier. Do you have more information so I can reproduce it?

skmp:
Latest android studio generated

```
    0x790234fc0c <+0>:  sub    sp, sp, #0x10             ; =0x10 
    0x790234fc10 <+4>:  str    w0, [sp, #0xc]
    0x790234fc14 <+8>:  str    x1, [sp]
```

Maybe it depends on the compiler version?

skmp:
(and/or optimization level?)

skmp:
Seems to be triggered by `-O0` or `-O1`

flyinghead:
That's it: `-O0` does create a prologue

PR #1725, opened 2019-08-11
cmake: misc changes

scribam:


skmp (commented):
libDL is a requirement for initializing opengl, and the libs themselves are not needed. 

Have you seen the updated makefiles? It is probably best yo model after it

scribam:
> libDL is a requirement for initializing opengl

libDL is still there with `${CMAKE_DL_LIBS}`

> the libs themselves are not needed.

Which libs are you referring to? Except for X11, none of them are marked as mandatory on Linux

skmp:
Glesv2 and libgl are resolved at runtime. libGL is still needed for GLX
though.

-D GLES is also not needed anymore (it does nothing). Both the gles/gl3 and
gl41 renderer are now supported at the same time now, with dynamic
detection between gles2, gles3, gl3 or gl4.

Also another gotcha is that the code doesn't support both EGL and GLX at
the same time.

(Initialization is a pita *sigh*)

scribam:
Got it, have a look to the latest commit and let me know if it is better.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5.0 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
```
skmp, line 1: Do we really use 3.10 features? Is it widely available - eg on odroids and such?

CMakeLists.txt
```diff
@@ -229,30 +225,84 @@ elseif (${HOST_OS} EQUAL ${OS_LINUX} OR ${HOST_OS} EQUAL ${OS_ANDROID})
...
+        link_libraries(${LIBUDEV_LIBRARIES})
+      endif()
+
+      pkg_check_modules(X11 REQUIRED x11)
+      if(X11_FOUND)
+        add_definitions(-DSUPPORT_X11)  ## don't use GLES ?
+        list(APPEND osd_SRCS ${d_core}/linux-dist/x11.cpp)
+        include_directories(${X11_INCLUDE_DIRS})
+        link_libraries(${X11_LIBRARIES})
+        if(OpenGL_GLX_FOUND)
```
skmp: I'd only fallback to GLX if EGL is not found. 

EGL can initialize both GLES and GL, GLX can only initialize GL.

EGL works over X11, afair, but it is a bit more fragile than GLX over older driver versions and such.

CMakeLists.txt
```diff
@@ -229,30 +225,84 @@ elseif (${HOST_OS} EQUAL ${OS_LINUX} OR ${HOST_OS} EQUAL ${OS_ANDROID})
...
     else()
       list(APPEND osd_SRCS 
         .//android-studio/reicast/src/main/jni/src/Android.cpp
         .//android-studio/reicast/src/main/jni/src/utils.cpp
     #   .//android-studio/reicast/src/main/jni/src/XperiaPlay.c
+        ${d_core}/utils/glinit/egl/egl.cpp
       )
+
+      add_definitions(-DUSE_EVDEV)
+      link_libraries(asound dl EGL Xext)
```
skmp: Aren't `asound`, `EGL` and `Xext` handled by the auto-detection logic?

libswirl/utils/glinit/glx/glx.cpp
```diff
@@ -3,6 +3,7 @@
 #include "utils/glwrap/gl3w.h"
 
 #include <GL/glx.h>
+#include <GL/glxext.h>
```
skmp: `glext` was not used on purpose, is there some reason to have it as a dependency?

skmp (changes requested):
I'm certainly not a cmake expert, comments are based on my partial understanding of it.

CMakeLists.txt
scribam: Version 3.10 is required to have the variables `OpenGL_GLX_FOUND` and `OpenGL_EGL_FOUND`. What is the operating system used by odroids? If it is Ubuntu 18.04 or later then it will be fine (https://packages.ubuntu.com/bionic/cmake).

CMakeLists.txt
scribam: What do you think about switching `SUPPORT_EGL` and `SUPPORT_GLX` in this case (and in the other places) to make EGL the default?
https://github.com/reicast/reicast-emulator/blob/d56d5bb9021205f2c25dc0f3e8b43b6f881186ff/libswirl/linux-dist/main.cpp#L356-L363

libswirl/utils/glinit/glx/glx.cpp
scribam: To remove this:
https://github.com/reicast/reicast-emulator/blob/6c440b51ffff03866d8e0eecd41316ad915df488/libswirl/utils/glinit/glx/glx.cpp#L87-L89

CMakeLists.txt
scribam: I have only tested my changes on Linux, this is the Android path. I don't know how Android works so I kept this line to not break things on this system.

libswirl/utils/glinit/glx/glx.cpp
skmp: The definitions are guaranteed to not change - this was on purpose to not depend on `glext.h`. Please do a git-rebase and undo this change

CMakeLists.txt
skmp: Android uses the standard makefile system, the cmake/android path is not used currently. CMake is used only for windows right now (before your changes, that is).

CMakeLists.txt
skmp: Ubuntu 18.04 is too recent - most of these board comes with distros from 2016 or so. We can keep them on the make path though, and only use CMake for newer distros

scribam: You can update CMake from the official site if needed but I understand your point. No reason to drop the Makefile right now and I am happy with the idea of using CMake for newer distros.

CMakeLists.txt
scribam: Not sure if my question was clear since there is no feedback here. I suggest to write the lines above like this so EGL will be the default if supported:
```
#if defined(SUPPORT_EGL)
 	return egl_Init(x11_win, x11_disp); 
#elif defined(SUPPORT_GLX)
 	return glx_Init(x11_win, x11_disp); 
#else 
 	#error "only x11 supported" 
 	return true; 
 #endif
```

CMakeLists.txt
scribam: Would it be better to remove the cmake/android path then? I don't mind to keep these lines as a hint for a potential Android support.

libswirl/utils/glinit/glx/glx.cpp
scribam: Change reverted, code rebased

CMakeLists.txt
skmp: `SUPPORT_GLX` also adds hooks in the x11 window creation - only one of them should be defined from the build system right now. See https://github.com/reicast/reicast-emulator/blob/d56d5bb9021205f2c25dc0f3e8b43b6f881186ff/libswirl/linux-dist/x11.cpp#L398-L413

CMakeLists.txt
skmp: Yes, it is best to remove it. We can re-add it in a clean way when things are ready for it

scribam: Right, android path removed

CMakeLists.txt
scribam: Done : GLX support added only if GLX is available and EGL is not.

sl1pkn07:
please add this line near line 305 in `reicast/cmake/config.cmake`

`set(_C_FLAGS "${_C_FLAGS} -D USES_HOMEDIR")`

this defined is called in the makefile but not in the cmakefiles

skmp (approved):

scribam:
- `USES_HOMEDIR` is only set in the linux Makefile:
https://github.com/reicast/reicast-emulator/blob/ad8378a4d3904fa1eacb38280727cd5f35542a6b/reicast/linux/Makefile#L30
- `USES_HOMEDIR` can only be found in the the `linux-dist/main.cpp` file which is included only for  `PLATFORM_LINUX`
https://github.com/reicast/reicast-emulator/blob/f205c0f53a821d31a4c39052b623e7393f96a20c/libswirl/core.mk#L103-L105
So, if I don't misunderstand, `USES_HOMEDIR` is always set when its value is checked. Is it still necessary to keep this flag?

sl1pkn07:
if not use `USES_HOMEDIR`, always when run reicast search the config in the directory where is launched, and create the files if not found. enable it always search and store the config in XDG home directory

without defined (in both cases builded from cmake)

~~~
└───╼  reicast
Warning: Unable to open the config file './emu.cfg' for reading (No such file or directory)
Creating new empty config file at './emu.cfg'
~~~

with defines `USED_HOMEDIR`

~~~
└───╼  reicast
Config dir is: /home/sl1pkn07/.config/reicast/
Data dir is:   /home/sl1pkn07/.local/share/reicast/
~~~

add `-DUSES_HOMEDIR` in the cmake command line rules not work, need add define in the cmake scripts

seems this also fixes if launch on macos (same XDG directory/file structure)

scribam:
@sl1pkn07 IMHO, the behavior when using `USES_HOMEDIR` should be the default and a flag should not be required to have it. That's why I would like to remove the `USES_HOMEDIR` references from the code.
Anyway, I am not going to do it in this PR, consider CMake support on Linux as WIP. Thank you for your feedback though.

PR #1726, opened 2019-08-13
input: Fix memleak in GamepadDevice class

scribam:


skmp:
On the road today, will review later at night :)

Is input mapper always created? I saw some NULL checks on the logic around
it

skmp:
Following a bit the code logic, it looks like  `input_mapper` is also stored in https://github.com/reicast/reicast-emulator/blob/dd18eaa1f029831058564f7a10b07bc3c6ebaca8/libswirl/input/mapping.cpp#L169 (in  `InputMapping::loaded_mappings`).

You probably want the cleanup to be done on shutdown there, and not on the per-gamepad class, as we'll end up with dangling deleted pointers.

Nice catch though, the code doesn't cleanup these instances

skmp (commented):
Change the logic to cleanup the leaks in `InputMappings` teardown

skmp (changes requested):
Change the logic to cleanup the leaks in `InputMappings` teardown

scribam:
I don't plan to work on this anymore.

PR #1728, opened 2019-08-14
reios: Remove extra printf argument

scribam:
There is only one `%X` in the format string or maybe `r[5]` is wanted and there is a `%X` missing?


skmp (approved):

skmp:
Looks like r5 isn't used for these

PR #1729, opened 2019-08-15
Minor misc changes

scribam:


skmp:
This pull request **fixes 4 alerts** when merging fc7b42706b6aa1b91ee11b546b6f6e46f3ec5002 into 4b1bbe0640fb126388ff98ee70d91de6a7ddc69e - [view on LGTM.com](https://lgtm.com/projects/g/reicast/reicast-emulator/rev/pr-be03d03577aa2763d7a15f4d3b23a417c0348957)

**fixed alerts:**

* 3 for Missing header guard
* 1 for Non\-virtual destructor in base class

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

libswirl/hw/maple/maple_cfg.cpp
```diff
@@ -124,8 +124,7 @@ bool maple_atomiswave_coin_chute(int slot)
 
 void mcfg_Create(MapleDeviceType type, u32 bus, u32 port, s32 player_num = -1)
 {
-	if (MapleDevices[bus][port] != NULL)
-		delete MapleDevices[bus][port];
+	delete MapleDevices[bus][port];
```
skmp, line 127: The null check here is intentional - there's no guarantee that the device has been instantiated before

libswirl/hw/maple/maple_devs.cpp
```diff
@@ -113,8 +113,7 @@ void maple_device::Setup(u32 prt)
 }
 maple_device::~maple_device()
 {
-	if (config)
-		delete config;
+	delete config;
```
skmp, line 116: It is likely this is intentional

libswirl/hw/naomi/naomi_cart.cpp
```diff
@@ -119,8 +119,7 @@ static bool naomi_LoadBios(const char *filename, Archive *child_archive, Archive
 		}
 	}
 
-	if (bios_archive != NULL)
-		delete bios_archive;
+	delete bios_archive;
```
skmp, line 122: It is likely this is intentional. Bios Archive might not always be found.

skmp (changes requested):
Most, if not all, the null checks are intentional, as objects might be missing, and `delete` on nullptr used to be illegal. I don't know if semantics have changed since then with C++11.

If you want to do this kind of cleanups, I'd recommend to use rewrite the logic to use smart pointers from `std` instead.

scribam:
According to https://en.cppreference.com/w/cpp/memory/new/operator_delete
> In all cases, if ptr is a null pointer, the standard library deallocation functions do nothing.

PR #1732, opened 2019-08-17
pulseaudio: Add missing return value to the "push" function

scribam:
Without this change, PulseAudio doesn't work for me.
Log before this change:
```
recSh4 Init
Using Interpreter
Sh4 Reset
Initializing audio backend "pulse" (PulseAudio)...
Using Interpreter
fault_handler: Blocking before restoring default SIGSEGV handler
PulseAudio: pa_simple_write() failed!
fault_handler: Blocking before restoring default SIGSEGV handler
fault_handler: Blocking before restoring default SIGSEGV handler
fault_handler: Blocking before restoring default SIGSEGV handler
```
Log after this change:
```
recSh4 Init
Using Interpreter
Sh4 Reset
Initializing audio backend "pulse" (PulseAudio)...
Using Interpreter
Invalid GD-DMA start, SB_GDEN=0.Ingoring it.
VREG = 03 ARMRST 00
VREG = 03 ARMRST 01
VREG = 03 ARMRST 00
Invalid GD-DMA start, SB_GDEN=0.Ingoring it.
VREG = 03 ARMRST 00
VREG = 03 ARMRST 01
VREG = 03 ARMRST 01
VREG = 03 ARMRST 00
Invalid GD-DMA start, SB_GDEN=0.Ingoring it.
VREG = 03 ARMRST 00
VREG = 03 ARMRST 01
VREG = 03 ARMRST 01
VREG = 03 ARMRST 00
Saved ./data/dc_nvmem.bin as nvmem
```

PR #1733, opened 2019-08-17
cmake: Rework OpenMP integration

scribam:


skmp:
@scribam I merged this one by accident, thought it was #1732. Looking at the changes, it does look alright thought. Any reason to revert?

scribam:
The changes are only for the Linux path, no changes expected for Windows so I consider these changes pretty safe.

Issue #1574, opened 2019-04-27
cmake: fail build in linux


sl1pkn07:
**Please Note: This form is the minimum required information for submitting bugs.**  
**Removing this form may lead to your issue being closed until it is completed.**

| Platform | Branch | Hash | CIDL |
| ------ | ------ | ---- | ---- |
| archlinux| master | ce90d43c344f19fdce915deef636ded933bc0d08 | GIT |

**_Description of the Issue_**

fail build in linux

**_Debugging Steps Tested_**

~~~
  cd build
  cmake ../reicast-emulator \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DUSE_QT=ON
~~~

**_Logs Gathered_**

```
[ 28%] Building CXX object CMakeFiles/reicast.dir/core/rend/gl4/gldraw.cpp.o
[...]

PR #1734, opened 2019-08-17
linux: Always check first in "HOME" / "XDG_CONFIG_HOME" for user config and data

scribam:
Remove `USES_HOMEDIR` from the code : `USES_HOMEDIR` is only defined in the Linux Makefile and the file `libswirl/linux-dist/main.cpp` is only compiled for Linux so I think it is safe to remove it.

### DIFF
--- a/libswirl/linux-dist/main.cpp
+++ b/libswirl/linux-dist/main.cpp
@@ -12,6 +12,7 @@
 #include <sys/param.h>
 #include <sys/mman.h>
 #include <sys/time.h>
+#include <sys/stat.h>
 #include "hw/sh4/dyna/blockmanager.h"
 #include "hw/maple/maple_cfg.h"
 #include <unistd.h>
@@ -33,10 +34,6 @@
 
 #if defined(USE_SDL)
 	#include "sdl/sdl.h"
-#endif
-
-#if defined(USES_HOMEDIR)
-	#include <sys/stat.h>
 #endif
 
 #if defined(USE_EVDEV)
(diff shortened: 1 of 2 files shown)


sl1pkn07:
why this is closed?

sl1pkn07:
reference issues:

https://github.com/reicast/reicast-emulator/issues/1574#issuecomment-521743963
https://github.com/reicast/reicast-emulator/issues/1574#issuecomment-521766386
https://github.com/reicast/reicast-emulator/pull/1725#issuecomment-521807239
https://github.com/reicast/reicast-emulator/pull/1725#issuecomment-522079767

skmp:
Good question, this looks like a nice fix. @scribam wana re-open this?

scribam:
Pull request reopened

skmp:
Okay, trying to understand how this fits in with the rest of the system

This PR effectively always makes it use HOME/XDG_HOME_* stuff, right?

I guess the cmake files need to be modified as well?

Did we ever /not/ define `USES_HOMEDIR`? I'm thinking we might break portable installations or downstream front-ends . 

I'll go over the relevant code later today and hopefully figure everything out, been in PVR land the past few weeks.

sl1pkn07:
cmake don' t have setted `USES_HOMEDIR`  (not need if merge this PR)

but Makefile in linux folder, yes

if is a problem with portable devices, then is better set it in the CMakeLists.txt as configurable option instead of this PR (or only apply a portion of him)

greetings

skmp:
Okay, after considering all cases I'd like us to have a `--portable` command line option to override the home dir. @sl1pkn07 would you be up to implement this, or shal I merge this and open a new PR myself?

sl1pkn07:
better if done by you. my code skills is more than poor (i can modify existent code, but not create new code)

greetings

skmp:
Err, I meant to ask @scribam

PR #1735, opened 2019-08-17
cmake: Misc platforms rework

scribam:


skmp:
Are there any benefits from switching from `_ANDROID` to `__ANDROID__`? We generally avoid "magic platform defines" as they often make things harder to figure out and customise -- vs defines we control ourselves.

libswirl/hw/pvr/ta_ctx.cpp
```diff
@@ -15,26 +15,28 @@ tad_context ta_tad;
...
-  }
-
-  *memptr = memalign(alignment, size);
-  if (*memptr == NULL) {
-    return errno;
-  }
-
-  return 0;
-}
+#ifdef __ANDROID__
```
skmp, line 18: Since you're touching this, can you move it to the android-specific platform code?

scribam:
For me, `_ANDROID` looks more "magic" than `__ANDROID__` which is part of the official tool aka the Android NDK. Using custom defines means you will have to declare them for every build system whereas `__ANDROID__` is available for everyone.

```
➜  ~ ~/android-ndk-r20/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi16-clang -x c /dev/null -dM -E | grep ANDROID
#define __ANDROID_API__ 16
#define __ANDROID__ 1
```

libswirl/hw/pvr/ta_ctx.cpp
scribam: Into `libswirl/android/Android.cpp`?

scribam:
@skmp A list of macros to write portable code is available here: https://blog.kowalczyk.info/article/j/guide-to-predefined-macros-in-c-compilers-gcc-clang-msvc-etc..html
I plan to remove the custom macros from reicast to switch to them but, first, I would like to know if it is something you will accept to merge.

skmp:
Current preference is to have platform detection complexity as part of the
build configuration, as much as possible.

I'd be opposed to change this without a good reason to do so.

My reasoning is that magic defines are much harder to grep for and keep
under control, and sometimes don't work as expected for cross
compiles/custom toolchains

scribam:
Thank you for the feedback. I have a different opinion but since you are the maintainer, it is up to you to decide what is best for your project.

PR #1751, opened 2019-10-13
Updated list of libraries required for Linux build and how to include Pulseaudio

dinoboy197:
Was building Reicast for Linux last night on Ubuntu 18.04 and found these extra dependencies were needed. I also wanted to ensure that Pulseaudio was used, so included the flag for those who want to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,10 +118,15 @@ Building for Linux
 	* libasound2-dev
 	* mesa-common-dev
 	* libgl1-mesa-dev
+	* libudev-dev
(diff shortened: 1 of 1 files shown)


skmp (approved):

skmp:
@dinoboy197  Took me a  while to merge, but better late than never, right?

dinoboy197:
I'm just happy that Reicast exists, so I'm happy to help out and contribute where I can!

PR #1753, opened 2019-10-24
Vertical scaling support added

mar753:
I have added vertical scaling support (SCALER_CTL), this fixes the issue e.g. in Crazy Taxi pause screen, where background is not properly aligned (see the red arrow):

### DIFF
--- a/libswirl/rend/gles/glestex.cpp
+++ b/libswirl/rend/gles/glestex.cpp
@@ -474,6 +474,18 @@ typedef map<u64,TextureCacheData>::iterator TexCacheIter;
 
 TextureCacheData *getTextureCacheData(TSP tsp, TCW tcw);
 
+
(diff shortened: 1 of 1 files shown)


skmp:
@mar753 can't the fraction be calculated like `0x0400/(float)0x0400`?

skmp:
(Sorry for the late review, been busy irl)

mar753:
First stage here is to convert binary fraction coded number (which Dreamcast is using in this case) to IEEE 754 - I have used `double` for better precision, which can make a difference here (`binaryFractionToDouble()` function). Second stage is to multiply current screen height by this factor, `round` is necessary to be used here, because `ceil`/`floor` or simple division are not sufficient here and may generate +/- 1px or more uncertainty. I have tested it very carefully with different games and rounding the final double number only, does generate graphics exactly like on Dreamcast, without any stretching, with the picture covering the whole screen.

PR #1760, opened 2019-12-31
OOPIFY the source code

skmp:


skmp:
@davidgfnet what do you think on the direction this is going?

*edit* also happy new year and all

skmp:
@flyinghead requesting your input as well, as you'd likely be affected if you try to rebase/cherry-pick from upstream in the future

Issue #1763, opened 2020-01-12
Create a virtual serial device while running (e.g /dev/ttyS0)


Kazade:
**_Description of the Issue_**

It would be great if Reicast would create a bi-directional serial device on boot so that devs can use dcload-serial to run their games. The advantage of this (rather than running directly) is we can use gdb debugging and also write files to the PC (e.g. profiling data).

Thoughts:

 - It should probably be enabled by default, as the Dreamcast has a serial device by default
 - The existing serial-to-terminal option should still exist, but if enabled output should go to both the terminal and the serial device
 - The device name should probably be overridable on the command line

**_Debugging Steps Tested_**

Not debugging steps, but proof it works:

 - Boot the dcload serial CDI image in Reicast
 - Note that a tty has been created in /dev
 - Upload a .elf file using dc-load-ser into Reicast with the `-c .` flag passed to mount the current directory

skmp:
@Kazade I'll look into it for late FEB/MARCH release

skmp:
#1792 Adds a new setting in Advanced, "Create Virtual Serial Port". When enabled reicast should output on init

```
[...]
Creating virtual serial port at /dev/pts/8
[...]
```
Note the device name is dynamic.

Executing
```dc-tool -x ~/Downloads/burritro2.elf  -t /dev/pts/8```

Should work as expected.
Gets around 200 kb/sec line speed right now, which isn't terrible all things considered.

skmp:
@kazade the only feature not implemented is being able to specify the file name. I'd rather pospone that til the rest of the code is clean enough. Is it alright if we open a follow up ticket for it?

skmp:
Also split #1793

skmp:
#1794 adds the file + buffered reading (7x faster!)

Also, for documentation, to enable all of the magic and boot dcload:

`./reicast -config config:Debug.VirtualSerialPortFile=ttyReicast -config config:Debug.VirtualSerialPort=yes ~/reicast-roms/dcload-103-serial-dj3.cdi 
`

skmp:
Closing this for now, kazade please re-open if you want more features :)

PR #1792, opened 2020-02-07
Linux: Virtual Serial Port support

skmp:
Implementing #1763 for @Kazade and other dc devs


skmp:
Implementing #1763 for @kazade and other dc devs

PR #1798, opened 2020-02-12
Add bios autoboot option

Cpasjuste:
This PR allow booting directly to the bios by giving the "bios" arg instead an image file.

### DIFF
--- a/libswirl/gui/gui.cpp
+++ b/libswirl/gui/gui.cpp
@@ -354,6 +354,8 @@ struct ReicastUI_impl : GUI {
             std::string game_file = cfgLoadStr("config", "image", "");
             if (!game_file.empty())
             {
+                if(game_file == "bios")
+                    game_file = "";
                 if (gui_start_game(game_file))
                     gui_state = Closed;
             }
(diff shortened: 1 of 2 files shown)


libswirl/gui/gui.cpp
```diff
@@ -354,6 +354,8 @@ struct ReicastUI_impl : GUI {
             std::string game_file = cfgLoadStr("config", "image", "");
             if (!game_file.empty())
             {
+                if(game_file == "bios")
```
skmp, line 357: Do you mind adjusting the logic in `libswirl.cpp::int GetFile(char* szFileName)`? You'll have to change a few empty strings as well in GUI, but it's a much cleaner solution to make "bios" mean "bios" at the core level. I never liked using an empty string anyway.

skmp (changes requested):

libswirl/gui/gui.cpp
skmp: Hmm, also maybe name it no "nodisk" to make it clear what it does?

baka0815 (changes requested):
@Cpasjuste you should add a parameter ``-bios``/``--bios`` (or ``-nodisk``/``--nodisk``) like in line 127 of cl.cpp instead of using "bios" as a magic value for a disk image.

skmp:
@baka0815 I don't mind using a special value for no disk, but I'd rather it be "nodisk" than "bios". Command line arguments aren't settable from the cfg file, that's why I try to keep them at a minimum

skmp:
functionality merged via #1818

PR #1807, opened 2020-02-15
Primitive Reference Renderer

skmp:
I've contacted nkeynes and they are chill for us to re-use `tacore` under GPL.

The reference renderer is implemented in two layers.

`struct refrend` implements the CORE interface and calls the actual implementation (`struct refsw` via `interface RefRendInterface`.

The idea is that later on, a `refcl` or `refgl` or `refhdl` implementation will follow.


libswirl/gui/gui_renderer.cpp
```diff
@@ -581,6 +581,9 @@ struct GUIRenderer_impl : GUIRenderer {
     virtual void QueueEmulatorFrame(std::function<bool()> cb) {
         callback_mutex.Lock();
         frameDone = false;
+        #if FEAT_TA == TA_LLE
```
skmp, line 584: These are here because slightly different timing / automagic frameskip between LLE and HLE

libswirl/hw/pvr/Renderer_if.cpp
```diff
@@ -205,12 +205,15 @@ static bool rend_frame(u8* vram, TA_context* ctx) {
         rend_init_renderer(vram);
     }
 
-    bool proc = renderer->Process(ctx);
-
-	if (!proc || !ctx->rend.isRTT) {
-		// If rendering to texture, continue locking until the frame is rendered
-		pend_rend = false;
-		re.Set();
+    bool proc = true;
```
skmp, line 208: These are here because slightly different timing / automagic frameskip between LLE and HLE

libswirl/hw/pvr/pvr_mem.cpp
```diff
@@ -103,7 +104,11 @@ void TAWrite(u32 address,u32* data,u32 count, u8* vram)
 	u32 address_w=address&0x1FFFFFF;//correct ?
 	if (address_w<0x800000)//TA poly
 	{
+		#if FEAT_TA == TA_HLE
```
skmp, line 107: Compile time toggle between backends - #define for zero perf cost for builds like emscripten

libswirl/gpl/lxdream/tacore.cpp
```diff
@@ -0,0 +1,1258 @@
...
+
+static void ta_end_list() {
+    if( ta_status.current_list_type != TA_LIST_NONE ) {
+        //asic_event( list_events[ta_status.current_list_type] );
+        asic_RaiseInterrupt(list_events[ta_status.current_list_type]);
+    }
+    ta_status.current_list_type = TA_LIST_NONE;
+    ta_status.current_vertex_type = TA_VERTEX_LISTLESS;
+    ta_status.poly_vertex_size = 0;
+    memset(&ta_status.poly_context[1], 0, 4);
```
snickerbockers: This is pedantic since most/all platforms do pin a char at being 1-byte, but using a hardcoded length is technically implementation-defined behavior because the memset function is specified in terms of chars instead of bytes.  sizeof(ta_status.poly_context[1]) is more compliant because it is also specified in terms of chars instead of bytes.

libswirl/gpl/lxdream/tacore.cpp
```diff
@@ -0,0 +1,1258 @@
...
+        plistpos -= ta_status.tilelist_size;
+    }
+    TA_NEXT_OPB = plistpos;
+    //MMIO_WRITE( PVR2, TA_LISTPOS, plistpos );
+    ta_status.tilelist_start = plistpos;
+    ta_status.polybuf_start = TA_ISP_BASE & 0x00F00000; // MMIO_READ( PVR2, TA_POLYBASE )
+}
+
+static uint32_t parse_float_colour( float a, float r, float g, float b ) {
+    int ai,ri,gi,bi;
```
snickerbockers, line 270: if you switch this to unsigned and then multiply by 255 instead of multiplying by 256 and subtracting 1, all the `< 0` branches can be removed.

libswirl/gpl/lxdream/tacore.cpp
```diff
@@ -0,0 +1,1258 @@
...
+            ta_status.clip.x2 = data[6].i & 0x3F;
+            ta_status.clip.y2 = data[7].i & 0x0F;
+            if( ta_status.clip.x2 >= ta_status.width )
+                ta_status.clip.x2 = ta_status.width - 1;
+            if( ta_status.clip.y2 >= ta_status.height )
+                ta_status.clip.y2 = ta_status.height - 1;
+            break;
+        case TA_CMD_POLYGON_CONTEXT:
+            if( ta_status.state == STATE_IDLE ) {
+                ta_init_list( TA_POLYCMD_LISTTYPE( data->i ) );
```
snickerbockers, line 1163: This is probably something that should be logged in the event that the list type changes without closing the list, AFAIK none of the docs say what should happen in this case.

libswirl/gpl/lxdream/tacore.cpp
```diff
@@ -0,0 +1,1258 @@
...
+                if( TA_IS_MODIFIER_LIST( ta_status.current_list_type ) ) {
+                    ta_parse_modifier_context(data);
+                } else {
+                    ta_parse_polygon_context(data);
+                }
+            }
+            break;
+        case TA_CMD_SPRITE_CONTEXT:
+            if( ta_status.state == STATE_IDLE ) {
+                ta_init_list( TA_POLYCMD_LISTTYPE( data->i ) );
```
snickerbockers, line 1180: same as above in the vertex context

snickerbockers (approved):
I don't see anything obviously wrong with this, so my comments are mostly just minor pedantry.  I do think the TAFIFO input should have some sort of optional logging for situations where games do seemingly contradictory things with no obvious correct behavior (including but not limited to changing the list type without opening a new list).

libswirl/gpl/lxdream/tacore.cpp
skmp: This would slightly modify the FLOAT -> INT curve though for values less than 127. I assume this has been cross tested with hardware, as there are TA tests in lxdream's source. I'll make a point to test it vs hardware and see if it can be simplified.

libswirl/gpl/lxdream/tacore.cpp
skmp: Will add warning. HLE already has one.

libswirl/gpl/lxdream/tacore.cpp
skmp: Will add warning. HLE already has one.

libswirl/gpl/lxdream/tacore.cpp
skmp: Will use `sizeof()`

skmp: Actually, I'll rewrite as `ta_status.poly_context[1] = 0;` - poly_context is u32 already.

libswirl/gpl/lxdream/tacore.cpp
skmp: Looking at the logic a bit closer, this happens under normal operation - it's a check for the per-list list-init, not the global one. Global list init is handled in `lxd_ta_init`

libswirl/gpl/lxdream/tacore.cpp
skmp: Looking at the logic a bit closer, this happens under normal operation - it's a check for the per-list list-init, not the global one. Global list init is handled in `lxd_ta_init`

PR #1812, opened 2020-02-25
Refrend: Add ReiDBG, RefSW fixes

skmp:
Screenshot from 2020-02-25 21-23-12
Screenshot from 2020-02-25 04-38-36
Screenshot from 2020-02-25 03-20-13


mrneo240 (commented):
Changeset removes magic numbers and replaces with clearer logic.
Also moves closer towards how hardware operates especially with filtering and coplanar drawing.

PR #1814, opened 2020-02-28
arm64: Fix CANONICAL_FALLBACK check

skmp:


skmp:
Tested it a bit and it seems to work fine

Issue #1813, opened 2020-02-27
wglSwapLayerBuffers cripples Intel integrated graphics video output on Windows 10


theurgus:
It is at it says on the tin. The latest alpha branch is unusable on my Intel Ivy Bridge integrated GPU as the menu takes literally 10 seconds to process events. Using the Very Sleepy profiler, I've tracked the problem to the wglSwapLayerBuffers function call causing the video driver's client library to wait for something through WaitForMultipleObjects. I'm not a graphics guy, so I've simply replaced the offending call with a plain SwapBuffers and the problem seems gone, though I don't know if it breaks anything else in the process.

skmp:
`wglSwapLayerBuffers` is the correct way to flip the screen, afaik. What kinds of speeds are you getting? Maybe it waits for vsync?

theurgus:
Well, it seems that a single frame gets rendered in 10 seconds or more. I can't tell you the exact numbers as I can't get past the menu to get to the FPS counter.

This is speculative, but from [what I've read](https://docs.microsoft.com/en-us/windows/win32/opengl/overlay--underlay--and-main-planes) on wglSwapLayerBuffers, the function has to do with overlays - you know, those abominations from the 90s where the GPUs only ever worked fullscreen, so you had to set your window's background to some punch-through color so that the GPU's output is visible? 

There's also [a remark](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-wglswaplayerbuffers#remarks) for the function itself: apparently, you have to do some feature checks before using it.

As for whether it's the correct way, I have no idea, but I've literally never seen this function before, as all books and tutorials I've seen don't bother using anything than SwapBuffers anyway =/

skmp:
Grepping the internet a bit more it seems that `SwapBuffers` should be the same as `wglSwapLayerBuffers(hdc, WGL_SWAP_MAIN_PLANE)`. Sadly, documentation does not seem very clear, not without knowing how the entire GDI stack works. 

I'm open to switching it if it fixes your problem - can you make a PR for it? Leave the `wglSwapLayerBuffers` commented out with a note about perf.

I'll test on my laptop later today and if it works fine I think we're safe to merge

theurgus:
[...]

PR #1815, opened 2020-02-28
Remove wglSwapLayerBuffers in favor of SwapBuffers

theurgus:
The change affects Windows graphics context management: the former function is relatively obscure and causes stuttering at least on Intel Ivy Bridge GT1 iGPU, while the latter is more or less standard, consistent with other platforms and works just fine.

@skmp,
I've finally managed to do the chore and I'm as confused as before. The issue only happens on my Intel Ivy Bridge GT1 iGPU, while the one year older Sandy Bridge GT1 works fine and GeForce GT630M works fine with either or both of functions. So, a driver bug?

I also ran a search on a couple of projects I remember: retroarch, redream, sdl, glfw, fltk, opengl superbible samples and nehe samples all use SwapBuffers and make no mention of the old one, except in systems headers.

All in all, the change seems safe enough to commit without further testing. I haven't left the old one as a comment either: with the function being unknown as it is, the chance of it resurfacing is nil.

The parent issue: https://github.com/reicast/reicast-emulator/issues/1813

### DIFF
--- a/libswirl/utils/glinit/wgl/wgl.cpp
+++ b/libswirl/utils/glinit/wgl/wgl.cpp
@@ -173,8 +173,7 @@ bool wgl_Init(void* hwnd, void* hdc)
 
 bool wgl_Swap()
 {
-	wglSwapLayerBuffers(ourWindowHandleToDeviceContext, WGL_SWAP_MAIN_PLANE);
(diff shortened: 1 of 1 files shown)


skmp (approved):
LGTM

skmp:
@theurgus please make sure you've read README.md#contributing and all :). Thanks for your PR!

PR #1822, opened 2020-03-05
Fixed the Mac version!

einsteinx2:
- Xcode project now builds and runs (macOS version only, haven't worked on iOS build)
- Fixed virtual serial port on macOS (tested with dcload-serial)
- Removed dependency on Swift (ported all Swift code to Objective-C) so that hopefully cmake alone can be used without Xcode in the future
- Refactored all of the macOS platform code, using more idiomatic Objective-C++ and Objective-C
- Tested and working great on my Mac with macOS Catalina 10.15.3 (19D76). Tested building on Linux and it compiled successfully, but couldn't run it as my VM doesn't have proper OpenGL support.

Known issues:
- When closing the app after BIOS or game is loaded it will either hang or crash, but can be force killed.
- When dragging the window between monitors with different DPIs (one retina, one not), the ImGUI scaling is not corrected, however games rescale properly.
- When resizing the window while displaying the GUI, there is a red color behind the black background that can be seen. This can also be seen after booting into the BIOS or a game, and resizing the window to be very tall. The top part will have black bars on the sides depending on the window's aspect ratio, but the bottom area will be all red. I'm not sure if this is something in libswirl or the macOS code's EmuGLView...
- Nothing else! Seems to work great :)


einsteinx2:
Closing this PR as I'll be pushing directly to this repo, will open a fresh PR.

PR #1825, opened 2020-03-05
Fixed the Mac version!

einsteinx2:
This new PR fixes the Android and Windows builds I broke.

- Xcode project now builds and runs (macOS version only, haven't worked on iOS build)
- Fixed virtual serial port on macOS (tested with dcload-serial)
- Removed dependency on Swift (ported all Swift code to Objective-C) so that hopefully cmake alone can be used without Xcode in the future
- Refactored all of the macOS platform code, using more idiomatic Objective-C++ and Objective-C
- Tested and working great on my Mac with macOS Catalina 10.15.3 (19D76). Tested building on Linux and it compiled successfully, but couldn't run it as my VM doesn't have proper OpenGL support.
- Added Azure CI for Linux and macOS (Big thanks to @sonic182 for the azure-pipeline.yml example!)

Known issues:
- When dragging the window between monitors with different DPIs (one retina, one not), the ImGUI scaling is not corrected, however games rescale properly.
- When resizing the window while displaying the GUI, there is a red color behind the black background that can be seen. This can also be seen after booting into the BIOS or a game, and resizing the window to be very tall. The top part will have black bars on the sides depending on the window's aspect ratio, but the bottom area will be all red. I'm not sure if this is something in libswirl or the macOS code's EmuGLView...
- Nothing else! Seems to work great :)


libswirl/deps/khronos/GL3/gl3w.h
```diff
@@ -1,5 +1,5 @@
 #pragma once
-#include "kronos/GL3/gl3.h"
+#include <GL3/gl3.h>
```
skmp: Is there some reason to use the system headers instead of the pre-packaged ones? We don't have the glheaders as a dependency right now.

libswirl/linux/common.cpp
```diff
@@ -51,8 +57,10 @@ void sigill_handler(int sn, siginfo_t * si, void *segfault_ctx) {
 	
     rei_host_context_t ctx;
     
-    context_from_segfault(&ctx, segfault_ctx);
-
+    // TODO: BEN fix this properly
```
skmp, line 60: What needs to be fixed here?

libswirl/linux/common.cpp
```diff
@@ -105,19 +113,21 @@ extern "C" u8* generic_fault_handler ()
...
 	        "movb $0, (%%eax)"
-		#elif HOST_CPU == CPU_X64
+		#elif HOST_CPU == CPU_X64 && HOST_OS == OS_DARWIN
 	        "movq %0, %%rsp\n"
-	        "call generic_fault_handler\n"
+	        "call _generic_fault_handler\n"
 	        "movb $0, (%%rax)"
+        #elif HOST_CPU == CPU_X64
+            "movq %0, %%rsp\n"
+            "call generic_fault_handler\n"
```
skmp: I'd prefer to have the `HOST_OS == OS_DARWIN` as a nested check here as the rest of the ASM is the same

libswirl/rend/gles/imgui_impl_opengl3.cpp
```diff
@@ -134,13 +134,15 @@ void    ImGui_ImplOpenGL3_RenderDrawData(ImDrawData* draw_data, bool save_backgr
     // Backup GL state
     glActiveTexture(GL_TEXTURE0);
     bool clip_origin_lower_left = true;
-
+    
+#if HOST_OS != OS_DARWIN
```
skmp, line 138: This is feels like a bit of a hack :/

reicast/apple/emulator-osx/emulator-osx/AppDelegate.mm
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// Called when the application terminates
+void os_gl_term() {
+    [_sharedInstance shutdownEmulator];
+}
+
+static void gl_resize() {
+    // Handle HDPI by multiplying the scale factor
+    rend_resize(_glViewSize.width * _backingScaleFactor, _glViewSize.height * _backingScaleFactor);
```
skmp, line 249: is this in pixels? I don't know how scaling works for mac

skmp (changes requested):
Can you please updated the license headers to follow the pattern as in the rest of the emulator? Looks good overall, I've left a few comments.

skmp:
Also, I think I have enabled azure pipelines, not sure.

libswirl/deps/khronos/GL3/gl3w.h
einsteinx2: Ah this was something I did early on when just trying to get it to compile. I think it was having issues with that header for some reason. I should be able to change it back now.

libswirl/linux/common.cpp
einsteinx2: I had meant to ask you about the correct usage here. The original line was `context_from_segfault(&ctx, segfault_ctx);` but the `context_from_segfault` function now only takes one parameter, so I had to change it to `context_from_segfault(&ctx);`. I wasn't sure what the `segfault_ctx` parameter was used for or if I broken anything by omitting it. Everything seems to work, but do you have any insight on why the second parameter was removed from the `context_from_segfault` function? Is it just no longer necessary?

libswirl/linux/common.cpp
einsteinx2: Sure no problem

libswirl/rend/gles/imgui_impl_opengl3.cpp
einsteinx2: I'll take another look at this one. I don't remember why I did it that way.

reicast/apple/emulator-osx/emulator-osx/AppDelegate.mm
einsteinx2: Yep this is in pixels. Normally everything is done with "points", which is part of the problem here. When initializing the Mac's GL view with a size of 640x480, its framebuffer gets set to 640x480 in points, which is actually 1280x960 in pixels when using a "retina" display aka a display with backing scale factor of 2.0. However the Reicast renderer is thinking only in pixels, so without this rend_resize call multiplying by the scale factor, the renderer output takes up only the bottom left quarter of the view on retina displays.

This just corrects for that and ensures that the framebuffer of the gl view and of the renderer are the same on high dpi displays.

libswirl/deps/khronos/GL3/gl3w.h
einsteinx2: Changed back, works fine, must have been something with the earlier xcode project config that was giving me problems.

libswirl/rend/gles/imgui_impl_opengl3.cpp
einsteinx2: So the problem here is that `glClipControl` and `GL_CLIP_ORIGIN` are undeclared identifiers on Mac, so I had used the #if block to remove them from the Mac build entirely.

einsteinx2:
@skmp Added the license headers, and resolved all of your requested changes except the ones I replied with questions to. Also unfortunately, doesn't look like it's using Azure CI. Maybe because it was enabled after this PR was created?

einsteinx2:
Fixed the crash on close issue after bios/game loading. Removed that from the known issues list.

libswirl/linux/common.cpp
skmp: It's no longer necessary - it keeps the context data structure internally now. Not the best design, but it seemed good at the time.

libswirl/rend/gles/imgui_impl_opengl3.cpp
skmp: That looks like a bug in our headers/glwrap. Can you add a comment about it?

skmp (approved):
Overall, LGTM. Can you remove the comment about the context - the current code is fine. Also, add a note for the GL functions so it's obvious why #ifdefs are needed there

skmp:
Merged. @einsteinx2 thanks for your first contribution!

PR #1830, opened 2020-03-13
WIP: Initial libretro port

DimitrisVlachos:
So I've being having some issues with the windows port.

First off all some important dll stubs are never called. Like retro_load_game where Im allowed to create an opengl context (or copy over the shared stubs of the main executable and patch my own vector stubs).

I tried different methods but none seem to work so far. The process is pretty straightforward.

To my surprise though most of the Core emulators don't work on windows with retroarch. For example Flycast exits abnormally with the error : Unable to load content.

I'm in debate whether I should check Linux ports or debug the behavior in Retroarch's source code .
@skmp


DimitrisVlachos:
Updated OP

skmp:
@DimitrisVlachos hum, I'd guess you'd best ask some of their devs about it. 

Switching to linux sounds good - we need to support them on android as well, and that is *nix based.

skmp:
@einsteinx2 can you lend a hand in implementing cmake the proper way in this?

@DimitrisVlachos we also need to fixup the CI once cmake is good

Otherwise looks good :)

Issue #1783, opened 2020-02-02
Show warning on first save-state use


skmp:
Warning should say something among the lines

"Save states are not guaranteed to be compatible between releases, please use with caution and always save to VMU"

with button "I Understand" or "Okay"

superfunc:
Looks like this can be closed out (handled by #1831)

PR #1831, opened 2020-03-14
Message on Save State (fix for issue #1783)

b10z:
This fix displays a popup message (as mentioned here #1783) the first time the Save State button is pressed. 
Tested on Android (!only!, i couldn't get visual studio build to work as expected for some reason) and works fine for me. if there is any issue please let me know :D

### DIFF
--- a/libswirl/types.h
+++ b/libswirl/types.h
@@ -360,8 +360,7 @@ using namespace std;
 #endif
 
 
-// NOTE: Always inline for macOS builds or it causes duplicate symbol linker errors
-#if DEBUG && HOST_OS != OS_DARWIN
+#if DEBUG
 //force
 #define INLINE
 //sugest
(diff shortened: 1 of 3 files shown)


libswirl/types.h
```diff
@@ -360,8 +360,7 @@ using namespace std;
 #endif
 
 
-// NOTE: Always inline for macOS builds or it causes duplicate symbol linker errors
-#if DEBUG && HOST_OS != OS_DARWIN
+#if DEBUG
```
skmp: Can you revert this?  it might break the osx build

skmp:
Looks good overall, can you please revert the #ifdef so that other platforms aren't affected?

libswirl/types.h
b10z: sure thing! i didn't even notice that change.. its fixed now i suppose! :D

skmp (approved):

skmp:
Thx for the contribution! 🏍️

Issue #1795, opened 2020-02-11
Add a "force landscape" option for android


skmp:
Some users seem to oddly like that behavior.

ghost:
Auto rotation?

skmp:
Implemented in alpha

PR #1834, opened 2020-03-17
Force orientation for Android

b10z:
As requested here #1795. This feature can be accessed by settings->general at any time, and the changes are applied instantly (no need for restart). There are three options there (force Portrait, force Landscape and Auto). Everything works fine on my Android device. If there is any issue, please let me know.  :)

### DIFF
--- a/libswirl/android/Android.cpp
+++ b/libswirl/android/Android.cpp
@@ -116,9 +116,10 @@ JNIEXPORT void JNICALL Java_com_reicast_emulator_emu_JNIdc_getControllers(JNIEnv
 
 JNIEXPORT void JNICALL Java_com_reicast_emulator_emu_JNIdc_setupMic(JNIEnv *env,jobject obj,jobject sip)  __attribute__((visibility("default")));
 
-SETTINGS_ACCESSORS(Nosound, aica.NoSound, jboolean)
-SETTINGS_ACCESSORS(Widescreen, rend.WideScreen, jboolean)
+SETTINGS_ACCESSORS(Nosound, aica.NoSound, jboolean);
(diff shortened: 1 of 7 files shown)


skmp (changes requested):
Can you make "auto" / the default value 0?

We generally try to make the settings in a way that "zero" is the default configuration

skmp:
Looks good overall

b10z:
> Can you make "auto" / the default value 0?
> 
> We generally try to make the settings in a way that "zero" is the default configuration

Ready! :D

skmp (approved):
Waiting for CI and then will merge

Tyfp:
Hi i just checked it out on ad7452a and the screen hasn't. Been auto rotated at all.

skmp:
By default it rotates to your phone. You have to set landscape from the
options if you want to force that

Tyfp:
I kinda did that on ad7452a but the options are olny showing rotate to 90 durgree on this lgk7 phone and nothing seems to work.

Tyfp:
Im at the settings now and it olny says rotate screen 90 even if i did that the screen will be too huge to look at therefore the screen hasn't been adjusted to size

Tyfp:
Tried the update one to it's still showing the same thing i need help on this please.

b10z:
hey @Tyfp ! 
Force Orientation option is under General tab in Settings. This fix has nothing to do with the option "Rotate screen 90" which is under Video tab. If you are still facing this problem, please let me know your device and android version.

Tyfp:
Oh ok i think i got it now thanks.

PR #1839, opened 2020-03-21
Fixed all warnings on Mac/Clang and Linux/GCC

einsteinx2:
Couldn't test Windows, but we'll see if the CI system has any issues.


skmp (approved):
LGTM

Issue #1840, opened 2020-03-22
Please add support for Redump.org UNLICENSED / MIL-CD (Multisessional CD-Rom) bin/cue


dizzzy77:
Hi reicast team, thank you very much for adding support for Redump style bin/cue for GD-Roms last year. It was very kind and thoughtful of your team <3

Mil-CDs and Unlicensed games (Multisessional CD-Roms) being a different format, are defined in their own way in our sets, example:

```
REM SESSION 01
FILE "GameShark Video Game Enhancer - Upgrade Disk (USA) (Unl) (Track 1).bin" BINARY
  TRACK 01 AUDIO
    INDEX 01 00:00:00
REM SESSION 02
FILE "GameShark Video Game Enhancer - Upgrade Disk (USA) (Unl) (Track 2).bin" BINARY
  TRACK 02 MODE2/2352
    INDEX 01 00:00:00
```

As TOSEC is also considering using the same-style cues as us when they add multisessional CD-Roms to their Dreamcast sets. So I figured now would be a good time to ebeg your team to add support.

If you wish to test with bins, a few examples are searchable archive.org: "GameShark CDX (USA) Dreamcast" and "Cheats 'N Codes Volume 1 (USA) Dreamcast"

Please consider adding support if time provides. Thank you for the consideration. Keep up the great work.

skmp:
Will try to fit in 20.04/20.04b.

Thanks for the heads up

skmp:
@DimitrisVlachos @stjordanis @einsteinx2 can any of you looking into extending our cue parser in imgread?

skmp:
Hum, might actually do it for today's live stream

skmp:
I'll do that

skmp:
@dizzzy77 is there some CUE spec I can review somewhere? I'll do some guesswork but hacky code is hacky

dizzzy77:
REM SESSION designates the session of the multisessional disc being called. As I understand both sessions need to be mounted similar to how you mount LD and HD areas in the Redump GD-Rom cues.

Below are the cues for all of the Dreamcast multisessional discs in redump:

```
REM SESSION 01
FILE "Action Replay CDX - Demo Version (UK) (Unl) (Track 1).bin" BINARY
  TRACK 01 AUDIO
    INDEX 01 00:00:00
REM SESSION 02
FILE "Action Replay CDX - Demo Version (UK) (Unl) (Track 2).bin" BINARY
[...]

PR #1841, opened 2020-03-22
WIP: Maybe parse a session based cue correctly?

skmp:
Partially address #1840


dizzzy77:
Might be a helpful clue? Unlicensed cheat/game discs have only redbook audio in SESSION 1, whereas Official Games have data as Track 1.

PR #1842, opened 2020-03-22
Cmake for macOS fully implemented including Mac app bundle creation!

einsteinx2:
- Included latest OpenMP static lib for macOS in the project (Apache license with LLVM exception)
- Fixed Xcode Project build configuration to remove the requirement for a separately installed LLVM, now builds with standard Apple supplied Clang!
- Removed MainMenu.xib and replaced with programmatically created menu bar to allow for easier cmake config
- Re-added credits to flyinghead in 2 macOS files
- Enabled FEAT_HAS_SOFTREND for clang (macOS only as it breaks Android builds, and soft rend “works” on macOS but seems _extremely_ slow...like waited a full minute and still didn’t get into the BIOS screen...)
- Fixed all warnings in macOS cmake build (there were a few new ones that didn't show up in Xcode)
- Removed unnecessary refsw_tsp_init.inl file from macOS app bundle Resources directory as it's not used from there
- Updated README with macOS cmake build instructions
- Added script to create Mac app bundle (Reicast.app) at the end of the cmake build process so it creates a fully compliant Mac app and not just a raw binary


einsteinx2:
Btw the inclusion of the OpenMP lib also applies to the Xcode build so now brew is completely unnecessary unless you prefer to use cmake over Xcode (unless you build the cmake tool from source in which you don’t need brew at all, but I would say 99% of Mac devs and power users have it anyway). 

The big win is not requiring a separate LLVM or libomp install anymore, especially since before it was using hard coded version numbers in the include paths which would break building any time libomp or LLVM was updated. 

Also now it should compile fine using cmake with just the ~300MB Apple command line dev tools package (install instructions in readme) for those that don’t want to download a ~7GB Xcode install just for this.

README.md
```diff
@@ -87,31 +87,36 @@ ant debug
...
+	* [Homebrew](https://brew.sh/)
+	* Cmake command line tool: `brew install cmake`
 
-From project root directory:
+- From project root directory:
 
-| iOS             | Mac                           |
-| --------------- | ----------------------------- |
-| `cd reicast-emulator/reicast/apple/emulator-ios` | `cd reicast-emulator/reicast/apple/emulator-osx` |
+`rm -rf build && mkdir build && cmake -S. -B./build && make -C build -j4`
```
skmp, line 115: does ninja also work?

reicast/apple/emulator-osx/emulator-osx/OSXKeyboardDevice.hpp
```diff
@@ -1,5 +1,6 @@
 /*
     This file is part of reicast-osx
+        original code by flyinghead
```
skmp, line 3: thx <3

skmp (approved):

PR #1849, opened 2020-03-27
fix(win32): also check height when saving window dimensions

Kamaropoulos:
Fixed small typo, `screen_width` was being checked twice instead of also checking `screen_height` when storing the window dimensions in the config file.


skmp (approved):

skmp:
hah

PR #1874, opened 2020-04-01
macOS console window with regex filtering

einsteinx2:
With Mac apps, it's a bit of a pain to view the console output of an app. You have to open the Console application--different from Terminal--and filter by your app name or you have to run the raw binary in Terminal which has other issues like missing images in the UI or missing libraries.

Many people don't know about this, so I added a "Show Console" option that allows you to view the redirected stderr/stdout in a window while still printing stderr/stdout normally (so it's viewable in the Console app or if you run the raw binary from Terminal).

reicast_macos_console_window


skmp:
Rebased on top of master, will merge if it passes CI

skmp (approved):
Looks good and clean!

PR #1886, opened 2020-04-07
import BBA emulation from nullDC

skmp:
- uses rtl driver from qemu
- seems to work with dc-load ip

TODO
- Fixup pcap backend


skmp:
Created followups #1890 #1889

PR #1894, opened 2020-04-13
Added basic file explorer functionality

b10z:
Hey guys!  Here i have some basic file explorer functionality (Delete file, rename file and bar on the top which allows to browse on specific folder on our content path. I attach a gif with the new interface adjustments.

(edit: new gif!) 
The code hasn't been fully tested on windows, since i could not make my keyboard to work correctly at any time with reicast for some reason. Some of the buttons that worked tho, were Crtl+V so that was enough for input checks, examples, etc but still not sure if everything works. 
Also, i tried to keep my code simple and not that hacky and i think that its ok most of the time. Any suggestion/opion is welcome ofc! 

Also, im almost done with archives. Sorry for the whole delay, but i realized that i needed threads to work a little too late and i had to write my code from scratch.  

Edit: Everything i mention bellow i think its done plus some changes to what i had done. Added some comments as well. :)  No bugs or errors for me on android 9, but havent test rename function on windows yet.

### DIFF
--- a/libswirl/android/Android.cpp
+++ b/libswirl/android/Android.cpp
@@ -646,6 +646,7 @@ static jmethodID VJoyStartEditingMID;
 static jmethodID VJoyStopEditingMID;
 static jmethodID VJoyResetEditingMID;
 static jmethodID MsgboxMID;
+static jmethodID showVirtualKeyboardMID;
 
 JNIEXPORT void JNICALL Java_com_reicast_emulator_BaseGLActivity_register(JNIEnv *env, jobject obj, jobject activity)
 {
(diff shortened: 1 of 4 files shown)


b10z:
Thinking about adding a new scrollable which includes all the files from multiple paths at once. Currently this code supports multiple paths but each path shows up only its own files and directories. (So you cant view content of /ReicastHome1/ and /ReicastHome2/ at once -you can do this separately from the bar- only if the path is like this/ReicastHome1/ReicastHome2). Also i believe i need to add a slash or something to each subfolder just to distinguish them from content paths used.

PR #1897, opened 2020-04-21
Android Archives(7z, zip) - with File explorer functionality

b10z:
Archived games support added for Android (i couldn't make a visual studio build, way too many errors to handle for now). 
Supported archives types: zip, 7zip.

Created using libzip and lzma, both libraries found under /deps. Long story short, i created a class where the whole decompress is happening and its called in a thread. Also, made the necessary adjustments to show archives like games, but in a diff category. If the comments are not enough, contact me clarify everything needed!

Some nice things i would also like to implement is a "Cancel" button, and maybe some more informations about the extracting file (like size, name, type, etc). 
 
ffgif

**EDIT 27/4
.GDI and .CUE files are supported as well. Reicast now creates a new directory if one these two types is found, and decompress everything there. This is used to prevent wrong files read between many games . 
Unfortunately, i do not have access to any CHD file to test. So i guess it works like CDI?

Also, i noticed that archives that have files inside folders, are not decompressing as they are supposed to. I'm working on a fix soon.**

Tested on two phones with Android 9 and everything works as expected.

### DIFF
--- a/libswirl/android/Android.cpp
+++ b/libswirl/android/Android.cpp
@@ -646,6 +646,7 @@ static jmethodID VJoyStartEditingMID;
 static jmethodID VJoyStopEditingMID;
 static jmethodID VJoyResetEditingMID;
 static jmethodID MsgboxMID;
+static jmethodID showVirtualKeyboardMID;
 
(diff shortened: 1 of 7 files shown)


Gamemulatorer:
> Tested on two phones with Android 9 and everything works as expected.

How about on android 11 scoped storage restrictions?

PR #1908, opened 2020-05-13
Fixed crash when updating controller settings

einsteinx2:
This was happening any time the emulator wasn't running yet. For example after first launching Reicast then setting up your controllers. From what I can tell, it would happen on all platforms, not just MacOS, though I've confirmed it's fixed there.

The actual fix is just the null check in the `maple_ReconnectDevices()` function. I added a similar null check in `maple_vblank()` just to be safe.


skmp:
We need to refactor a bit some stuff so these corner cases don't happen. Someday.

skmp (approved):

PR #1919, opened 2020-08-23
Add support for newer Raspberry Pi 4 with 8 GB RAM, BCM2711

danboid:
Add support for newer Raspberry Pi 4 with 8 GB RAM, the BCM2711.

I have only tested building reicast under Raspbian Buster but it works fine with this patch.


skmp:
Thanks!

